PR #1, opened 2015-09-17
First rendition of design doc

ExtraGravity:


bseto:
I think this is pretty good. Something we can implement pretty easily I think.

DesignDoc.adoc
```diff
@@ -0,0 +1,69 @@
...
+The theme of the game will be similar to http://old.casualcollective.com/#games/The_Space_Game[the space game] by Casual Collective. Although the gameplay will be very differnet. The player will only see a radar so the graphics will be very simple, mcuh of the view will be the player's "terminal". Think basic shapes on a radar. I was thinking circle is the main character, triangles are enemies, squares are projectiles.
+
+== Controls
+
+All controls will be on the keyboard, even in the start screen.
+
+In gameplay you will select your weapon via hotkeys 1-5. Once a weapon has been selected by pressing a hotkey, the interface will switch to targeting mode where you type the names of enemies to target them, every weapon will be unique, and enemies will also do various things.
+
+== Weapons
+
```
zacharykoo, line 22: I think there should be a weapon that is like 1 use, where it is for emergency and things gets too hard, there can be a weapon to neutralize things once, for people that are not that good on typing

beaesguerra:
if player is a circle, will it shoot from different sides of the circle? (do the enemies just come from one direction?)
can you move the player around or is it static? 

is the in-game currency based on points or just completing the level? is there a points system? 

what will be determine the end of a level? after a certain amount of enemies/words have passed or a time limit?

DesignDoc.adoc
ExtraGravity, line 22: I'll add that additions to the design doc, but I won't change it in this one because it's not a "problem". These pull requests are more for discussing problems rather than new ideas.

ExtraGravity:
<i>if player is a circle, will it shoot from different sides of the circle? (do the enemies just come from one direction?)</i>
It's a radar, to the player will simply be represented as a circle. Where things are shot from are completely decided by the weapon system.

<i>can you move the player around or is it static?</i>
Line 5 : "There is no movement control, like a Touhou type game where the ship is always moving forward"
I will add a commit to link gameplay of a touhou game. Here is a [video](https://www.youtube.com/watch?v=WdoF0iayjvs) for now.  Our game will have zero movement. Like in FTL where you have no control over your movement, but you use your systems effectively to not take damage. Here is a video of [FTL](https://www.youtube.com/watch?v=Acwjfkc_M90), gameplay at 23:00.

<i>is the in-game currency based on points or just completing the level? is there a points system?</i>
Kind of a mix of both, you'll gain some points for completing a level, but you also gain points for doing things in the mission well. Ideally, I'd want everything to be gained via completing levels, but to get all upgrades that's too many levels.

<i>what will be determine the end of a level? after a certain amount of enemies/words have passed or a time limit?</i>
That's a design decision for later, it could be time, it could be defeating a boss. Hard to say.

You said you saw typos earlier, write them in so I can fix them.

PR #5, opened 2015-09-20
Added difficulty section

zacharykoo:


ExtraGravity:
We won't include this in the prototype, we we'll put this on hold for the actual.

PR #6, opened 2015-09-21
Word generator

bseto:


wordgenerator.py
```diff
@@ -0,0 +1,106 @@
...
+    word_array = []
+    number_of_lengths = 0
+    smallest_length = []
+
+    def __init__(self):
+        with open(CONST_FILENAME) as infile:
+            for line in infile:
+                self.word_array.append(line.rstrip('\n'))
+        random.shuffle(self.word_array) 
+        self.word_array.sort(key=len)
```
ExtraGravity, line 26: So in this line, did you test that if they have the same len they'll still be random?

bseto, line 26: Line 24? Yes they will still be random after you sort them by length.

wordgenerator.py
```diff
@@ -0,0 +1,110 @@
...
+        for i in range(0, len(self.word_array)):
+            print("Index: " + repr(i) + " Word: " + self.word_array[i])
+
+
+if __name__ == '__main__':
+    tester = WordGenerator()
+    for i in range(0,100):
+        print("random word: " + tester.request_word(19,1))
+
+
```
ExtraGravity: Should remove the extra white lines at the end here, small nitpicky thing. I just finished fixing whitespace in the other files so it would be nice if this one conformed also.

ExtraGravity:
I'd like to request that this pull request also involves integration into master. So merge master into this branch before we merge the pull request.

Basically, after we merge the pull request, the target_tags should be generated by this class. So you'll need to change some code in the other files as well to integrate this class.

wordgenerator.py
```diff
@@ -0,0 +1,110 @@
...
+    '''
+    Will populate the list index_locations on where each new length starts
+    For example, index_locations[3] will give you the location of the first word with length 3
+    index_locations[4] will give you location of first word with length 4, etc.
+    index_prime_locations will be incremented and used to keep track of which words have been given out.
+    '''
+    def split_list(self, list):
+        self.number_of_lengths = len(list[len(list)-1]) - len(list[0]) 
+        self.smallest_length = len(list[0])
```
ExtraGravity, line 38: The only place `self.smallest_length` is used, is in this function.
Should just be a variable rather than a class member.

wordgenerator.py
```diff
@@ -0,0 +1,110 @@
...
+        self.split_list(self.word_array)
+
+    '''
+    Will populate the list index_locations on where each new length starts
+    For example, index_locations[3] will give you the location of the first word with length 3
+    index_locations[4] will give you location of first word with length 4, etc.
+    index_prime_locations will be incremented and used to keep track of which words have been given out.
+    '''
+    def split_list(self, list):
+        self.number_of_lengths = len(list[len(list)-1]) - len(list[0])
```
ExtraGravity: `len(list[len(list)-1]) - len(list[0])`  
This line is slightly cryptic. 

You also never use `self.number_of_lengths` outside of this function so I don't think it should be a member variable.

Since you calculate `self.smallest_length` anyways I'd suggest replacing these two lines with.

```
shortest_word_length = len(list[0])
longest_word_length = len(list[-1]) # [-1] is the pythonic way of getting the last item in a list
number_of_lengths = longest_word_length = shortest_word_length 
```

I think `shortest_word_length` is a better (but only slightly better) name than smallest_length.

wordgenerator.py
```diff
@@ -0,0 +1,110 @@
...
+    Will populate the list index_locations on where each new length starts
+    For example, index_locations[3] will give you the location of the first word with length 3
+    index_locations[4] will give you location of first word with length 4, etc.
+    index_prime_locations will be incremented and used to keep track of which words have been given out.
+    '''
+    def split_list(self, list):
+        self.number_of_lengths = len(list[len(list)-1]) - len(list[0]) 
+        self.smallest_length = len(list[0])
+        #For lengths that are smaller than the smallest word, we point them to index 0 in the array
+        for x in range (0, self.smallest_length):
```
ExtraGravity: `range(n) == range(0, n)`
Please shorten the range function calls.

wordgenerator.py
bseto, line 38: Ya, was planning on using it to check for words requests that are too small. With this current dictionary, smallest word size is 1 so this isn't really going to be used. However, if we have a dictionary with smallest word size being 3 or 4, and you request size 1. I was going to have it give you an error. Right now it will just give you whatever is in element 0.

wordgenerator.py
```diff
@@ -0,0 +1,110 @@
+import pygame
+import random
+import sys
+import math
+from copy import deepcopy
+
+CONST_FILENAME = "american-english.txt"
```
ExtraGravity: The all caps I think should imply that this is a CONST variable, I believe that's why nothing like `const` is included in python. `FILENAME` should suffice for this variable name.

wordgenerator.py
bseto: Won't be implementing the len(list[:-1]) as this is done with my c++ knowledge, and it will be easier to port if I just keep it as c++ as possible.

number_of_lengths is used often near the bottom

ExtraGravity: My bad, I was checking for the recurrences for `number_of_lengths` locally on my computer and I hadn't pulled the last commit. 

I still think it should be coded len(list[:-1]), the intention is more obvious. You might not be the one porting this back to c++.

wordgenerator.py
```diff
@@ -0,0 +1,110 @@
...
+import math
+from copy import deepcopy
+
+CONST_FILENAME = "american-english.txt"  
+
+'''
+Opens the file CONST_FILENAME and reads entire file to word_array list. Shuffles the array and then sorts according to length
+'''
+class WordGenerator():
+    index_locations = []
```
ExtraGravity: Not sure if you want to change this but I think it'd be good for you to know.

Writing the member variables out of any function without a `self` makes it a class member, not an instance member. Class members are similar to static member variables in c++. 

Initializing instance members should be done in the `__init__` function like:

```
def __init__(self):
    self.index_locations = []
    ...
```

I also think there should be some indicator that they are member variables, like a trailing underscore or a leading m_, I've been using a trailing underscore if you want to be consistent with the rest of the code.

wordgenerator.py
bseto: Not sure what len(list[:-1]) does, but when I checked the number_of_lengths value I got from doing it that way^, I got random numbers in the thousands.

google'd, maybe you meant  list[-1]?

wordgenerator.py
```diff
@@ -0,0 +1,110 @@
...
+            return False
+        if index == 0:
+            return True
+        return len(self.word_array[index]) != len(self.word_array[index-1])
+
+    '''
+    If you want words with length 4-6 (inclusive), request_word(4,3)
+    If you want words with length of 2, request_word(2,1)
+    '''
+    def request_word(self, index, number_of_words):
```
ExtraGravity: Reading from line 70-73 I'm confused at the parameter names. I don't feel like `index` and `number_of_words` really reflect the use case from the above comments. Looks more like `shortest_length` and `length_range` from the comments.

bseto: ya, shortest length would be a good variable. I was working on this for a while and in my head, index = length of word now

wordgenerator.py
bseto: I didn't like how I couldn't see what variables I had globally lol

wordgenerator.py
ExtraGravity: Yea you're right, it's `list[-1]`, `list[:-1]` is the list except for the last element, my bad :P
Updated original comment

ExtraGravity:
```
Traceback (most recent call last):
  File "terminalfighter.py", line 17, in <module>
    gamemaster = GameMaster(universe)
  File "/home/enoch/ext/TerminalFighter/gamemaster.py", line 15, in __init__
    self.spawn_main_character()
  File "/home/enoch/ext/TerminalFighter/gamemaster.py", line 21, in spawn_main_character
    the_main_character = MainCharacter(starting_position, self.universe_)
  File "/home/enoch/ext/TerminalFighter/maincharacter.py", line 24, in __init__
    self.weapons_ = [Rifle(self.universe_), HomingMissiles(self.universe_)]
  File "/home/enoch/ext/TerminalFighter/rifle.py", line 97, in __init__
    self.targeting_system = RifleTargetingSystem(universe)
  File "/home/enoch/ext/TerminalFighter/rifle.py", line 21, in __init__
    self.word_generator = WordGenerator()
  File "/home/enoch/ext/TerminalFighter/wordgenerator.py", line 25, in __init__
    self.split_list(self.word_array)
  File "/home/enoch/ext/TerminalFighter/wordgenerator.py", line 42, in split_list
    self.index_locations.append(self.locate_first_length(x))
  File "/home/enoch/ext/TerminalFighter/wordgenerator.py", line 53, in locate_first_length
    if self.check_if_first(index, requested_length):
  File "/home/enoch/ext/TerminalFighter/wordgenerator.py", line 65, in check_if_first
    if len(self.word_array[index]) != requested_length:
TypeError: list indices must be integers, not float
```

Getting this error with python2

ExtraGravity:
Still need to test the "looping back to beginning" feature, otherwise looks good.

bseto:
You can test it by using the words of length like 20 and 21. They only have a small amount of words with that length

PR #8, opened 2015-09-21
Targeting terminal class created and included terminal class into rifle.py

gordon-cheung:


targeting_terminal.py
```diff
@@ -0,0 +1,38 @@
...
+            if event.type == pygame.KEYDOWN:
+                if event.key == EVENT_KEY_ENTER:
+                    self.current_text_ = ""
+                if event.key == EVENT_KEY_BACKSPACE:
+                    self.current_text_ = self.current_text_[:-1]
+                elif event.key in range(EVENT_KEY_a, EVENT_KEY_z) and textWidth < 380:
+                    self.current_text_ += event.unicode.lower()
+
+    def draw_terminal(self, screen):
+        terminal_rect = pygame.Rect(10*self.DRAWING_SCALE_, screen.get_height()-(40*self.DRAWING_SCALE_), 400* self.DRAWING_SCALE_, 25*self.DRAWING_SCALE_)
```
bseto: Since this is now a single class that is always instantiated, would it be more wise to do this calculation once on `__init__` and then keep re-using `terminal_rect` @cheunggg @enochtsang?

screen is not passed in on **init**, but maybe you can keep a bool that makes it do this calculation only on the first time calling "draw_terminal"?  

EDIT: Just realized this is not a single class. However, this optimization I think still makes sense.

targeting_terminal.py
```diff
@@ -0,0 +1,38 @@
...
+class TargetingTerminal():
+      
+    def __init__(self, DRAWING_SCALE):
+        self.current_text_ = ""
+        self.ui_font_ = pygame.font.SysFont("monospace", 10*DRAWING_SCALE)
+        self.DRAWING_SCALE_ = DRAWING_SCALE
+
+    def update(self , events):
+        for event in events:
+            textWidth = self.ui_font_.size(self.current_text_)[0]
```
bseto: Optimization: 
Place the:
`textWidth = self.ui_font_.size(self.current_text_)[0]`
after the
`if event.type == pygame.KEYDOWN:`
Since this textWidth value would only be evaluated if there was a `pygame.KEYDOWN` event, calculate it only when that happens

targeting_terminal.py
ExtraGravity: I agree with Byron, `terminal_rect` should be a member variable initialized in `__init__`
Although calculating Rects are very quick so this isn't much of an optimization, in terms of coding style it still makes more sense to calculate this once in the `__init__`

bseto: Ya its not too much of an optimization, but every little thing counts in a game. This would have been called on every update on every targetting system (weapon).

ExtraGravity:
Can you increase the terminal font size in this pull request as well?

targeting_terminal.py
ExtraGravity: Also, one of the reasons I wanted `terminal` moved to it's own class is because I wrote a lot of magic numbers and it would get cluttered in the targeting system class to have labels for the magic numbers. Can you remove the constants (except for `/ 2` ones) to a member variable?

targeting_terminal.py
```diff
@@ -0,0 +1,38 @@
...
+                    self.current_text_ += event.unicode.lower()
+
+    def draw_terminal(self, screen):
+        terminal_rect = pygame.Rect(10*self.DRAWING_SCALE_, screen.get_height()-(40*self.DRAWING_SCALE_), 400* self.DRAWING_SCALE_, 25*self.DRAWING_SCALE_)
+
+        pygame.draw.rect(screen, BLUE, terminal_rect)
+        pygame.draw.rect(screen, WHITE, terminal_rect, 2*self.DRAWING_SCALE_)
+    
+        text_label = self.ui_font_.render(self.current_text_, 1, WHITE)
+        screen.blit(text_label, (15 * self.DRAWING_SCALE_, screen.get_height()-(30 * self.DRAWING_SCALE_)))
```
ExtraGravity: Magic numbers here as well

ExtraGravity: `screen.get_height()-(30 * self.DRAWING_SCALE_)`
The spacing around the operators is strange, I'd suggest:
`screen.get_height() - (30*self.DRAWING_SCALE_)`

bseto:
Ya I'd like the font to be larger too. Its variable atm with the Drawing Scale, but I think the base number of 15 is not large enough 
`self.ui_font_ = pygame.font.SysFont("monospace", 15*DRAWING_SCALE)`
https://github.com/enochtsang/TerminalFighter/blob/terminal/homing_missiles.py#L25

Make it a little bigger :)

ExtraGravity:
Currently the font size in is 10, in my opinion 15 is fine. Byron if you want it to be bigger I'd suggest making a pull request changing the font sizes across the whole program.

bseto:
I think it looks good :)

targeting_terminal.py
```diff
@@ -0,0 +1,52 @@
...
+
+# CONSTANTS
+BLACK = 0, 0, 0
+BLUE = 0, 0, 255
+WHITE = 255, 255, 255
+EVENT_KEY_a = 97
+EVENT_KEY_z = 122
+EVENT_KEY_ENTER = 13
+EVENT_KEY_BACKSPACE = 8
+fontSize = 15
```
ExtraGravity: The constants are inherently part of the TargetingTerminal class, they should be member variables.

targeting_terminal.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def update(self , events):
+        for event in events:
+            if event.type == pygame.KEYDOWN:
+                textWidth = self.ui_font_.size(self.current_text_)[0]
+                if event.key == EVENT_KEY_ENTER:
+                    self.current_text_ = ""
+                if event.key == EVENT_KEY_BACKSPACE:
+                    self.current_text_ = self.current_text_[:-1]
+                elif event.key in range(EVENT_KEY_a, EVENT_KEY_z) and textWidth < maxTextWidth:
```
ExtraGravity: You missed multiplying `maxTextWidth` with `DRAWING_SCALE`, as a result when using a scaling of 2, the max width of the terminal is not utilized.

targeting_terminal.py
```diff
@@ -0,0 +1,52 @@
...
+BLUE = 0, 0, 255
+WHITE = 255, 255, 255
+EVENT_KEY_a = 97
+EVENT_KEY_z = 122
+EVENT_KEY_ENTER = 13
+EVENT_KEY_BACKSPACE = 8
+fontSize = 15
+rectX = 10
+rectY = 40
+rectWidth = 400
```
ExtraGravity: Instead of using `rect` I'd suggest using `terminal`, `rect` could be the rect for anything.

ExtraGravity:
Looks good, merging

PR #9, opened 2015-09-22
Started RifleProjectile class. Added the ids_for_target_tags dictiona…

beaesguerra:
…ry. Determined the enemy ID of a target tag that matched the user input in the terminal. Determined the position of the targetted enemy.

### DIFF
--- a/rifle.py
+++ b/rifle.py
@@ -24,17 +24,24 @@ class RifleTargetingSystem():
         self.enemy_color_ = RED
         self.main_character_color_ = GREEN
         self.target_tags_ = dict()
+
+        self.ids_for_target_tags_ = dict() 
+
         self.current_text_ = ""
 
     def update(self, events):
         for enemy in self.universe_.enemies():
             if enemy.ID_ not in self.target_tags_:
-                self.target_tags_[enemy.ID_] = self.new_word()
+            	new_word = self.new_word()
+            	self.target_tags_[enemy.ID_] = new_word
+            	self.ids_for_target_tags_[new_word] = enemy.ID_ 
 
         for event in events:
             if event.type == pygame.KEYDOWN:
                 if event.key == EVENT_KEY_ENTER:
                     self.current_text_ = ""
+                    # ******** this is where you would update rifle projectile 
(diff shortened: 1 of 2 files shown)


rifle.py
```diff
@@ -87,6 +94,26 @@ def draw_terminal(self, screen):
...
+		target_ID = getTargetID(input_)
+		if (targetID): 
+			target_position = self.universe.enemies_[target_ID].position_ 	
+			print(target_position)	#test 
+			# shoot projectile towards where enemy was at that time. 
+
+	def getTargetID(input): 
+		for enemy in self.universe_.enemies(): 
+			target_tag_word = self.target_tags_[enemy.ID_]
+			if(input == target_tag_word)
```
bseto, line 113: I get an error here. The end of your "if" should have a ":"
Fix: if(input == target_tag_word):

rifle.py
```diff
@@ -87,6 +94,26 @@ def draw_terminal(self, screen):
...
+				return ids_for_target_tags_[input]
+			else		# input does not exist
```
bseto, line 115: **else** should also end with a colon
`else:`

rifle.py
ExtraGravity, line 113: Also, if statement in python don't need bracket, remove them for better python code :)

rifle.py
```diff
@@ -87,6 +94,26 @@ def draw_terminal(self, screen):
...
+			print(target_position)	#test 
+			# shoot projectile towards where enemy was at that time. 
+
+	def getTargetID(input): 
+		for enemy in self.universe_.enemies(): 
+			target_tag_word = self.target_tags_[enemy.ID_]
+			if(input == target_tag_word)
+				return ids_for_target_tags_[input]
+			else		# input does not exist 
+				return 0
```
ExtraGravity, line 116: I'd recommend returning `None` instead of 0, because having gameobjects of ID 0 may be a possibility later. Gameobjects will never had an ID of `None` so it's probably safer.

ExtraGravity:
On hold until feature finished

PR #10, opened 2015-09-22
Collision detection

ExtraGravity:


ExtraGravity:
I totally did not branch off of master, I will redo

PR #13, opened 2015-09-24
wrote damage access functions

ExtraGravity:


basic_grunt.py
```diff
@@ -36,3 +37,9 @@ def collision_box(self):
     	return pygame.Rect(self.position_[0]-self.size_/2,
         			       self.position_[1]-self.size_/2,
         			       self.size_, self.size_)
+
+    def take_damage(self, damage):
```
bseto, line 52: While you merge conflicts. Maybe make an error checker to make sure "damage" coming in is a positive number and not a negative

ExtraGravity, line 52: done

bseto:
Merging Pull request

PR #14, opened 2015-09-24
Optimized collisions

ExtraGravity:
We'll keep this branch for any collision optimizations we want to try out, so don't delete this branch after merging. 

As a note, I wrote that "main_character" doesn't care what it collides with because even if it does, it will not call any functions on what it collided with. This is enemies also.

On the other hand, if a enemy projectile collides with the main character, it will call a function "take_damage", so projectiles do care what they collide with.

### DIFF
--- a/universe.py
+++ b/universe.py
@@ -13,7 +13,8 @@ class Universe():
         self.gameobjects_ = dict()
         self.main_character_ = None
         self.enemies_ = dict()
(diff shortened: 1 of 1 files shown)


bseto:
Looks good. Is consistent with what we discussed at your house. Will merge after conflicts are fixed by @enochtsang

bseto:
Merging

PR #15, opened 2015-09-24
Code cleanup

ExtraGravity:


homing_missiles.py
```diff
@@ -1,40 +1,64 @@
...
+
+    """
+    Update Functions
+    """
 
     def update(self, events):
         for enemy in self.universe_.enemies_.values():
             if enemy.ID_ not in self.target_tags_:
-                self.target_tags_[enemy.ID_] = self.new_word()
+                self.target_tags_[enemy.ID_] = \
```
bseto, line 53: Lol wow thats a cool trick with "\". You are using it to escape the space right?

homing_missiles.py
```diff
@@ -58,20 +82,17 @@ def draw_entities(self, screen):
...
+                target_tag_word, self.text_antialias_, self.enemy_color_)
             width = self.ui_font_.size(target_tag_word)[0]
-            screen.blit(target_tag_label, 
+            screen.blit(target_tag_label,
                         (enemy.position_[0] * self.DRAWING_SCALE_ - (width/2),
-                        (enemy.position_[1] * self.DRAWING_SCALE_ + enemy.size_/2) + (5 * self.DRAWING_SCALE_)))
-
-    def new_word(self):
-        return self.word_generator_.request_word(12,1)
+                        (enemy.position_[1] * self.DRAWING_SCALE_ + enemy.size_/2) + (self.target_tag_y_spacing_ * self.DRAWING_SCALE_)))
```
bseto, line 95: Why wasn't this new-lined?

homing_missiles.py
```diff
@@ -1,40 +1,64 @@
...
+
+        self.current_text_ = ""
         self.enemy_color_ = YELLOW
+        self.font_size_ = 15
         self.main_character_color_ = GREEN
         self.target_tags_ = dict()
-        self.current_text_ = ""
+        self.target_tag_y_spacing_ = 5
         self.targeting_terminal_ = TargetingTerminal(DRAWING_SCALE)
+        self.text_antialias_ = 1
```
bseto, line 40: Out of curiosity. Are we using antialias in the proper context here? Like are we actually antialiasing anything?

bseto:
Other than the small comments above. Everything seems to work.

homing_missiles.py
ExtraGravity, line 53: Yup, allows you to continue on to the next line.

homing_missiles.py
ExtraGravity, line 95: Didn't know how to write it without confusing a reader. 
`(enemy.position_[0] * self.DRAWING_SCALE_ - (width/2),`
 `(enemy.position_[1] * self.DRAWING_SCALE_ + enemy.size_/2) + \`
`(self.target_tag_y_spacing_ * self.DRAWING_SCALE_)))`

I was thinking this would confuse the reader into there being 3 parameters.

bseto, line 95: Ah I guess that's a valid point. I was thinking that myself but wasn't sure if that was the reason. Ya keep it as it is now, I think it makes more sense if you don't make it new-lined

homing_missiles.py
ExtraGravity, line 40: https://www.pygame.org/docs/ref/font.html#pygame.font.Font.render 
From the docs, the second argument is antialias, to be honest I don't know much about what the "1" does though.

bseto:
Merging

Issue #19, opened 2015-09-28
'z' key button not working


beaesguerra:
When the 'z' key is pressed, it does not show up on the terminal on the game screen. The "current_text_" gets updated and adds the 'z' though. All the other letters show up on the terminal on the game screen.

PR #20, opened 2015-09-28
fixed issue #19

ExtraGravity:


gordon-cheung:
Would it not be better to change Event_Key_z = 123 from Even Key_z = 122?

ExtraGravity:
Because `EVENT_KEY_Z` is actually 122, that'd be false to list it as 123.

gordon-cheung:
Ok I understand, my mistake, good work on the fix!

PR #21, opened 2015-09-28
Completed Rifle Projectile and Added Temporary Lists for Game Objects in Universe

beaesguerra:
Altered the universe by placing created game objects (main character, enemies, enemy/friendly projectiles) into a temporary list before updating the universe. This fixed the error of changing the size of dictionaries while it was being iterated.


menu.py
```diff
@@ -0,0 +1,31 @@
+import pygame 
+
+import sys
```
ExtraGravity: The order of imports should be, 
built-in
3rd party
your own

So in this case, it should be.
`import sys`
`import pygame`
`from gamemaster mport GameMaster`

ExtraGravity:
The audio file is still in this pull request. `menu.py` should be removed also.

rifle.py
```diff
@@ -1,8 +1,9 @@
 import pygame
+import math
 
 from targeting_terminal import TargetingTerminal
 from wordgenerator import WordGenerator
-
+from gameobject import GameObject
 pygame.font.init()
```
ExtraGravity, line 9: imports should have a line separating them from all other code.

rifle.py
```diff
@@ -1,8 +1,9 @@
 import pygame
+import math
```
ExtraGravity, line 1: built-in modules should be above everything else.

rifle.py
```diff
@@ -25,9 +26,12 @@ def __init__(self, universe, DRAWING_SCALE):
 
         self.current_text_ = ""
         self.enemy_color_ = RED
+        self.projectile_color_ = WHITE
```
ExtraGravity: These member variables should be in alphabetical order.

rifle.py
```diff
@@ -36,17 +40,49 @@ def __init__(self, universe, DRAWING_SCALE):
         self.word_length_min_ = 3
         self.word_length_range = 3
 
-
+    def getTargetID(self,input):
```
ExtraGravity: In PEP8, methods are also with underscores. This should be 
`def get_target_id(self, input):`

rifle.py
```diff
@@ -36,17 +40,49 @@ def __init__(self, universe, DRAWING_SCALE):
         self.word_length_min_ = 3
         self.word_length_range = 3
 
-
+    def getTargetID(self,input): 
+        for enemy in self.universe_.enemies(): 
+            target_tag_word = self.target_tags_[enemy.ID_]
```
ExtraGravity: `target_tag_word` is only used in the next if statement, so should be fine to just put these 2 lines together. 
As in:
`if input == self.target_tags_[enemy.ID_]:`

rifle.py
```diff
@@ -36,17 +40,49 @@ def __init__(self, universe, DRAWING_SCALE):
...
+            if(input == target_tag_word):
```
ExtraGravity: if statement do not need brackets in python, for good style they should not be included.

rifle.py
ExtraGravity: `input` is a keyword in python, you should not use it as a parameter/variable name.

rifle.py
```diff
@@ -36,17 +40,49 @@ def __init__(self, universe, DRAWING_SCALE):
...
+                target_ID = self.ids_for_target_tags_[input]
+                break
+            else:       # input does not exist
```
ExtraGravity: this comment is not necessary

rifle.py
```diff
@@ -36,17 +40,49 @@ def __init__(self, universe, DRAWING_SCALE):
...
+        for enemy in self.universe_.enemies(): 
+            target_tag_word = self.target_tags_[enemy.ID_]
+            if(input == target_tag_word):
+                target_ID = self.ids_for_target_tags_[input]
+                break
+            else:       # input does not exist 
+                target_ID = -1
+        return target_ID  
+
+    def checkIfInputMatchesTargetTag(self):
```
ExtraGravity: method name should be underscores, not camelcase

rifle.py
```diff
@@ -36,17 +40,49 @@ def __init__(self, universe, DRAWING_SCALE):
...
     Update Functions
     """
-
     def update(self, events):
         for enemy in self.universe_.enemies():
             if enemy.ID_ not in self.target_tags_:
-                self.target_tags_[enemy.ID_] = \
-                            self.word_generator_.request_word(self.word_length_min_, self.word_length_range)
-
+                new_word = self.word_generator_.request_word(self.word_length_min_, self.word_length_range)
```
ExtraGravity: I like how this part is coded, good job!

rifle.py
```diff
@@ -36,17 +40,49 @@ def __init__(self, universe, DRAWING_SCALE):
...
 
-
+    def getTargetID(self,input): 
+        for enemy in self.universe_.enemies(): 
+            target_tag_word = self.target_tags_[enemy.ID_]
+            if(input == target_tag_word):
+                target_ID = self.ids_for_target_tags_[input]
+                break
+            else:       # input does not exist 
+                target_ID = -1
```
ExtraGravity: In Python, `None` is an explicit object, you should use `None` instead of `False` or `-1` in cases like these. In C++, this would be completely acceptable though :)

rifle.py
```diff
@@ -36,17 +40,49 @@ def __init__(self, universe, DRAWING_SCALE):
...
+        return target_ID  
+
+    def checkIfInputMatchesTargetTag(self): 
+        target_ID =  self.getTargetID(self.current_text_)
+        if (target_ID != -1): 
+            target_position = self.universe_.enemies_[target_ID].position_  
+            return target_position 
+        else:
+            return -1
```
ExtraGravity: Same thing as line 50, use `None`

rifle.py
```diff
@@ -36,17 +40,49 @@ def __init__(self, universe, DRAWING_SCALE):
...
+                new_word = self.word_generator_.request_word(self.word_length_min_, self.word_length_range)
+                self.target_tags_[enemy.ID_] = new_word
+                self.ids_for_target_tags_[new_word] = enemy.ID_ 
+
+        for event in events:
+            if event.type == pygame.KEYDOWN:
+                if event.key == EVENT_KEY_ENTER:
+                    target_location = self.checkIfInputMatchesTargetTag()
+                    self.current_text_ = ""
+                    if (target_location == -1):
```
ExtraGravity: As commented above, if you return None, this line could be written as:
`if not target_location:`
And again, if statements don't need parentheses.

rifle.py
```diff
@@ -36,17 +40,49 @@ def __init__(self, universe, DRAWING_SCALE):
...
+                        pass
```
ExtraGravity: In general, if statements should be not be empty. This `if else` block could be better written as:
`if target_location != -1:`
`your code here`

rifle.py
```diff
@@ -36,17 +40,49 @@ def __init__(self, universe, DRAWING_SCALE):
...
+                    else:
+                        rifle_projectile = RifleProjectile(self.universe_.main_character_.position_, target_location)
```
ExtraGravity, line 78: Other parts of the code used exactly as intended! Good job :)

rifle.py
```diff
@@ -58,6 +94,7 @@ def draw(self, screen):
         self.draw_entities(screen)
         self.draw_target_tags(screen)
         self.targeting_terminal_.draw_terminal(screen)
+        self.draw_friendly_projectiles(screen)
```
ExtraGravity: `draw_friendly_projectiles()` should be called in `draw_entities()`

rifle.py
```diff
@@ -88,6 +125,55 @@ def draw_target_tags(self, screen):
...
+                                          main_character.size_ * self.DRAWING_SCALE_,
+                                          main_character.size_ * self.DRAWING_SCALE_)
+        pygame.draw.rect(screen, self.main_character_color_, main_character_rect)
+
+class RifleProjectile(GameObject): 
+
+    def __init__(self, initial_position, target_position):
+        self.ID_ = self.create_ID()
+        self.initial_position_ = initial_position
+        self.position_ = self.initial_position_
```
ExtraGravity: As coding style, I know `self.initial_position_` and `initial_position` are the same, but in general you should use parameters where possible instead of doing a `self` call.

rifle.py
```diff
@@ -88,6 +125,55 @@ def draw_target_tags(self, screen):
...
+        self.target_position_ = target_position
+        self.speed_ = 15
+        self.size_ = 5
+
+    def update(self,events):
+        y2 = self.target_position_[1]
```
ExtraGravity: This math will always be the same, since it's done with `self.target_position_` and `self.initial_position_`. This should be done in the init method. Probably not directly in `__init__()` though, maybe have a function called `calculate_trajectory()`.

rifle.py
```diff
@@ -88,6 +125,55 @@ def draw_target_tags(self, screen):
...
+        x2 = self.target_position_[0]
+        x1 = self.initial_position_[0]
+
+        tx = x2 - x1
+        ty = y2 - y1
+        dist = math.sqrt(tx*tx+ty*ty)
+        rad = math.atan2(ty,tx)
+        angle = rad/math.pi*180
+
+        velx = (tx/dist)*self.speed_
```
ExtraGravity: `velx` and `vely` should have their full names, abbreviations are frowned upon for variables names. I'd suggest `x_velocity` and `y_velocity`. The x and y go in the front because more specific descriptors should be in ahead of less specific descriptors.

rifle.py
```diff
@@ -88,6 +125,55 @@ def draw_target_tags(self, screen):
...
+        vely = (ty/dist)*self.speed_
+
+        self.position_ = (self.position_[0] + velx, self.position_[1] - abs(vely))
```
ExtraGravity: This should be the only line in the `update()` method.

universe.py
```diff
@@ -45,30 +49,47 @@ def main_character(self):
...
     def create_enemy(self, the_enemy):
-        self.enemies_[the_enemy.ID_] = the_enemy
-        self.gameobjects_[the_enemy.ID_] = the_enemy
-
+        self.created_enemies.append(the_enemy)
+        
     def create_enemy_projectile(self, the_projectile):
-        self.enemy_projectiles_[the_projectile.ID_] = the_projectile
-        self.gameobjects_[the_projectile.ID_] = the_projectile
+        self.created_enemy_projectiles.append(the_projectilee)
```
ExtraGravity: typo here, `the_projectilee`

universe.py
```diff
@@ -45,30 +49,47 @@ def main_character(self):
...
         # print(str(len(self.gameobjects_)) + " gameobjects")
         # print(str(len(self.enemies_)) + " enemies")
 
     def create_enemy(self, the_enemy):
-        self.enemies_[the_enemy.ID_] = the_enemy
-        self.gameobjects_[the_enemy.ID_] = the_enemy
-
+        self.created_enemies.append(the_enemy)
+        
     def create_enemy_projectile(self, the_projectile):
```
ExtraGravity: Don't think it'd hurt to be more specific here, having the parameter be `the_enemy_projectile` instead of just `the_projectile`

ExtraGravity:
@bseto Please review also.

ExtraGravity:
`targeting_terminal` is is not really used except for drawing in this pull. Both the `targeting_terminal` and `targeting_system` and implementing the same `current_word` system.  

Currently how the projectile system is implemented is quite clean IMO and do not want to change it much. What I suggest is moving the `current_word` functionality away from terminal.  So instead in `targeting_terminal`'s drawing function, it gets passed in the current word from the targeting system. Additionally, the max size of the word will be decided in `targeting_system` and passed into the `__init__()` method of the `targeting_terminal`.

@bseto @cheunggg Thoughts?

bseto:
Okay wait. Right now targeting_terminal is used only for drawing? I do not like how rifle.py has it's own 
  `for event in events:`
           `if event.type == pygame.KEYDOWN:`

The update inside targeting_terminal should be the thing checking for "checkIfInputMatchesTarget". 

EDIT:

My bad, CheckIfInputMatchesTarget should be kept on the Targeting System. However, the targeting_terminal should return a word if it detects "EVENT_KEY_ENTER" so we don't need the same thing on TargetingSystem's update

Example: Inside rifle.py, RifleTargetingSystem, `update` function.

`self.current_text_ = self.targeting_terminal_.update(events)`
`targeting_terminal.update(events)` should pass either "" or a string. If its "", it means no one pressed "EVENT_KEY_ENTER", if its not empty, then someone pressed the "EVENT_KEY_ENTER"

That way, you can do:
`self.current_text_ = self.targeting_terminal_.update(events)`
`if self.current_text != "":`
`checkIfInputMatchesTargetTag`

@enochtsang  Thoughts?

ExtraGravity:
Right, so the targeting terminal is used only for drawing in this pull request. The original intention was to have it tell targeting system the target, but after seeing this implementation I wouldn't mind that being changed.  Changed so that the targeting terminal does not have its own events passed in, it gets current word passed in and displays that. So targeting  system, not terminal,  is keeping track of the typed buffer.

bseto:
I don't think that's good. It means every Targeting system will have a:
`for event in events:` checking for keyboard typing instead of having 1 class that has that logic.

ExtraGravity:
That functionality could be abstracted into an inherited function, for targeting system

bseto:
Okay so in the real implementation of C++, the TargetingSystem will be a superclass that things inherit then right? If so, then I can agree with this implementation

ExtraGravity:
Ok then, Bea please implement what we've just discussed.

rifle.py
```diff
@@ -26,28 +27,61 @@ def __init__(self, universe, DRAWING_SCALE):
         self.current_text_ = ""
         self.enemy_color_ = RED
         self.font_size_ = 15
+        self.ids_for_target_tags_ = dict() 
         self.main_character_color_ = GREEN
-        self.target_tags_ = dict()
+        self.projectile_color_ = WHITE
```
enochtsang: This initialization should be in alphabetical order.

rifle.py
```diff
@@ -26,28 +27,61 @@ def __init__(self, universe, DRAWING_SCALE):
...
-        self.word_length_range = 3
-
-
+        self.word_length_range_ = 3
+        
+    def get_target_id(self,terminal_input): 
+        for enemy in self.universe_.enemies(): 
+            if terminal_input == self.target_tags_[enemy.ID_]:
+                target_ID = self.ids_for_target_tags_[terminal_input]
+                break
```
ExtraGravity, line 49: This is good, it guarantees only one target will be shot in the extreme case that two enemies show up with the same target tag.

rifle.py
```diff
@@ -26,28 +27,61 @@ def __init__(self, universe, DRAWING_SCALE):
...
+            else:       
+                target_ID = None
+        return target_ID  
+
+    def check_if_input_matches_target_tag(self): 
+        target_ID =  self.get_target_id(self.current_text_)
+        if (target_ID != None):
```
ExtraGravity: The pythonic way to write this line would be:
`if target_ID:`

rifle.py
```diff
@@ -26,28 +27,61 @@ def __init__(self, universe, DRAWING_SCALE):
...
         self.text_antialias_ = 1
         self.ui_font_ = pygame.font.SysFont("monospace", self.font_size_*DRAWING_SCALE)
         self.word_generator_ = WordGenerator()
         self.word_length_min_ = 3
-        self.word_length_range = 3
-
-
+        self.word_length_range_ = 3
+        
+    def get_target_id(self,terminal_input):
```
ExtraGravity: parameters should have a space, so:
`(self, terminal_input):`

rifle.py
```diff
@@ -26,28 +27,61 @@ def __init__(self, universe, DRAWING_SCALE):
...
         self.word_length_min_ = 3
-        self.word_length_range = 3
-
-
+        self.word_length_range_ = 3
+        
+    def get_target_id(self,terminal_input): 
+        for enemy in self.universe_.enemies(): 
+            if terminal_input == self.target_tags_[enemy.ID_]:
+                target_ID = self.ids_for_target_tags_[terminal_input]
```
ExtraGravity, line 48: I'm noticing this now and I made this mistake in earlier code as well, abbreviations should not be capitalized in variable names, so `target_ID` should be `target_id`. We won't worry about fixing this in the prototype though, but so you know.

rifle.py
```diff
@@ -26,28 +27,61 @@ def __init__(self, universe, DRAWING_SCALE):
...
+                break
+            else:       
+                target_ID = None
+        return target_ID  
+
+    def check_if_input_matches_target_tag(self):
```
ExtraGravity: This method name with `check` implies that the return is a bool, but this is not the case. I would suggest `get_target_position(self, target_tag)` and have `self.current_text_` get passed in for the `target_tag` argument.

rifle.py
```diff
@@ -26,28 +27,61 @@ def __init__(self, universe, DRAWING_SCALE):
...
+                new_word = self.word_generator_.request_word(self.word_length_min_, self.word_length_range_)
+                self.target_tags_[enemy.ID_] = new_word
+                self.ids_for_target_tags_[new_word] = enemy.ID_ 
+
+        for event in events:
+            if event.type == pygame.KEYDOWN:
+                if event.key == EVENT_KEY_ENTER:
+                    target_location = self.check_if_input_matches_target_tag()
+                    self.current_text_ = ""
+                    if target_location:
```
ExtraGravity, line 77: Good, pythonic way of checking if `target_location` is `None`.

rifle.py
```diff
@@ -26,28 +27,61 @@ def __init__(self, universe, DRAWING_SCALE):
...
+                        rifle_projectile = RifleProjectile(self.universe_.main_character_.position_, target_location)
+                        self.universe_.create_friendly_projectile(rifle_projectile)
+                    
+                if event.key == EVENT_KEY_BACKSPACE:
+                    self.current_text_ = self.current_text_[:-1]
+                elif event.key in range(EVENT_KEY_a, EVENT_KEY_z+1):
+                    if len(self.current_text_)<= self.targeting_terminal_.max_word_size_:
+                        self.current_text_ += event.unicode.lower()
```
ExtraGravity: This line was updated in previous code to be:
`self.current_text_ += chr(event.key)` to fix a bug. It didn't get caught in the merge due to the fix being in a separate file, do make sure this bug fix gets in though.

rifle.py
```diff
@@ -88,6 +123,56 @@ def draw_target_tags(self, screen):
...
+class RifleProjectile(GameObject): 
+
+    def __init__(self, initial_position, target_position):
+        self.ID_ = self.create_ID()
+        self.position_ = initial_position
+        self.speed_ = 15
+        self.size_ = 5
+        self.velocity_ = self.calculate_trajectory(initial_position, target_position)
+
+    def calculate_trajectory(self, initial_position, target_position):
```
ExtraGravity, line 174: Much better function name than I was thinking of, beautifully named :)

rifle.py
```diff
@@ -88,6 +123,56 @@ def draw_target_tags(self, screen):
...
+        y2 = target_position[1]
+        y1 = initial_position[1]
+        x2 = target_position[0]
+        x1 = initial_position[0]
+
+        tx = x2 - x1
+        ty = y2 - y1
+        distance = math.sqrt(tx*tx+ty*ty)
+        rad = math.atan2(ty,tx)
```
ExtraGravity: Trigonometry is not needed to calculate the x and y velocity. Doing it this way also requires you to have the funky
`- abs(self.velocity_[1])` code on line 167. 

Copy pasted from `basic_grunt.py`:

```
        main_character_position = self.universe_.main_character().position_
        x_distance = main_character_position[0] - self.position_[0]
        y_distance = main_character_position[1] - self.position_[1]
        distance = (x_distance**2 + y_distance**2)**(1/2)
        ...
        x_velocity = (x_distance * self.speed_) / distance
        y_velocity = (y_distance * self.speed_) / distance
        self.position_ = (self.position_[0]+x_velocity, self.position_[1]+y_velocity)
```

Adapt it for this section of code.

ExtraGravity:
Looks good! Good job getting a difficult pull request finished guys :)

PR #23, opened 2015-09-29
Destroy listener

bseto:
Merged with Bea's Projectile_System branch
Added the destroy_listener.py. 
Game detects deaths and destroys objects now.


ExtraGravity:
Because this pull request is dependent on the rifle projectile pull request, we should wait for that one to be merged in first, remerge master into this one then look into finishing up this pull request.

basic_grunt.py
```diff
@@ -55,5 +56,20 @@ def take_damage(self, damage):
             print("Disregarding non positive damage")
         elif damage >= self.health_:
             self.health_ = 0
+            self.report_destroyed()
```
ExtraGravity, line 58: Since in this statement, object will definitely be destroyed. Is it even worth it to have `self.health_ = 0` anymore? Alternatively, we could put this report_destroyed functionality only once in the code, at the bottom of the update. So at the end of this object's update function, it will have 
`if self.health <= 0:`
`self.report_destroyed()`

destroy_listener.py
```diff
@@ -0,0 +1,13 @@
...
+"""
+    If your class wants to listen to when an object says it is destroyed
+    inherit this "interface" and then pass yourself in as an argument to the
+    register function
+"""
+
+class DestroyListener():
+
+    def reported_destroyed(self, type_gameobject):
+        raise NotImplementedError()
```
ExtraGravity, line 11: Didn't know about this in Python, nice very useful.

menu.py
```diff
@@ -0,0 +1,31 @@
+import pygame 
+
+import sys
```
ExtraGravity: Import ports should be in the order:
built in modules
third party modules
your own modules

So the `import sys` should be above the `import pygame`

ExtraGravity:
menu.py is included in this branch, it should be removed.

enochtsang:
So the bug fix for the 'z' button seems to have gotten lost somewhere. I'll push a fix for it.

enochtsang:
So I think what happened is that you didn't update your local master before merging master into this branch. Probably gonna need to remerge.

rifle.py
```diff
@@ -92,6 +92,7 @@ def update(self, events):
 
     def draw(self, screen):
         self.draw_background(screen)
+        #self.draw_friendly_projectiles(screen)
```
ExtraGravity: What was the historic reason for this?

rifle.py
```diff
@@ -181,7 +184,16 @@ def calculate_trajectory(self, initial_position, target_position):
 
         return (x_velocity, y_velocity)
 
+    def check_collisions(self):
+        collisions = self.universe_.get_collisions(self)
+        for collision in collisions:
+            collision.take_damage(1000)
```
ExtraGravity: This line is lightly misleading, that a collision would take damage? I think having `for gameobject in collisions:` would be clearer.

rifle.py
bseto: Yes. This was probably the most confusing merge conflict I ever dealt with and I was very confused whether or not that was used. I guess not xD?

PR #24, opened 2015-09-30
Added increasing difficulty (spawn rate increases) over time

gordon-cheung:


gamemaster.py
```diff
@@ -18,6 +18,17 @@ def __init__(self, universe, DRAWING_SCALE):
         self.main_character_spawn_height_ = self.universe_.height_*0.9
         self.spawn_main_character()
 
+        # Enter the spawn rate (grunts / second) at a specific time t
```
ExtraGravity, line 21: Remove these comments

ExtraGravity:
This code is difficult to understand and is not in line with the current code. And relies on magic numbers everywhere. Redo it.

PR #25, opened 2015-09-30
Spawn time increases over time

gordon-cheung:


ExtraGravity:
The spawn difficulty was originally 20, the minimum spawn delay was reached before the second enemy was even spawned. I changed it to 0.1 and takes about 20 enemies before reaching the minimum spawn delay. Otherwise looks good, merging good job.

PR #27, opened 2015-09-30
drew a grid to make the map feel more like a radar

ExtraGravity:


gordon-cheung:
Nice! Before merging pull request, will homing missiles have the same grid or a different one?

ExtraGravity:
This is up to the weapon, so it can be different, but I think they will be the same for the prototype.

PR #28, opened 2015-09-30
Cleaned variable names for targetting terminal class

gordon-cheung:


bseto:
I like the new names better than the old ones! good :)

targeting_terminal.py
```diff
@@ -16,22 +16,27 @@
...
-        self.fontSize_ = 15
-        self.maxTextWidth_ = 380
         self.terminal_rect_ = None
-        self.terminalHeight_ = 25
-        self.terminalRectX_ = 10
-        self.terminalRectY_ = 40
-        self.terminalWidth_ = 400
+        
+        # Terminal Visual Constants
+        self.terminal_RectX_ = 10
```
ExtraGravity: This is like a mix of capCase and under_scores. Should be consistent, so
`self.terminal_Rectx_` should be `self.terminal_rect_x_

ExtraGravity:
They are better overall, the naming has improved, but small things like the consistent use of underscores needs to be improved.

gordon-cheung:
Pliz Review again

ExtraGravity:
Looks good, merge in from master first. Make sure everything is still the same, then we can merge in the pull request.

EDIT: Nevermind looks good, merging.

PR #29, opened 2015-10-01
New not so basic grunt

bseto:


basic_grunt.py
```diff
@@ -66,6 +66,7 @@ def __init__(self, starting_position, universe):
         self.speed_ = 0.5
         self.health_ = 30
         self.universe_ = universe
+        self.enemy_type_ = "BasicGrunt"
```
ExtraGravity, line 71: I think `self.type_` would suffice in this setting

gamemaster.py
```diff
@@ -3,6 +3,7 @@
 import pygame
 
 from basic_grunt import BasicGrunt
+from not_so_basic_grunt import NotSoBasicGrunt
```
ExtraGravity: Imports should be in alphabetical order

gamemaster.py
```diff
@@ -37,6 +38,11 @@ def spawn_basic_grunt(self):
         the_basic_grunt = BasicGrunt(starting_position, self.universe_)
         self.universe_.create_enemy(the_basic_grunt)
 
+    def spawn_not_so_basic_grunt(self):
+        starting_position = [self.universe_.width_*random.choice(self.basic_grunt_x_spawn_locations), 0]
```
ExtraGravity: Instead of using `self.basic_grunt_x_spawn_locations` for `not_so_basic_grunt` we should either generalize the spawn location variable name, or make specific one for not so basic grunt. I think the first solution is better.

gamemaster.py
```diff
@@ -44,6 +50,7 @@ def update(self, events):
             if self.basic_grunt_spawn_delay_ < self.minimum_spawn_delay_:
                 self.basic_grunt_spawn_delay_ = self.minimum_spawn_delay_
         if self.basic_grunt_spawn_timer_ > self.basic_grunt_spawn_delay_:
+            self.spawn_not_so_basic_grunt()
```
ExtraGravity: I don't like how they're being spawned together, but I don't have any easy solutions either. It'll be messy to add another spawn timer (might be okay for prototype), if we do that the correct solution would be create a spawner class.

homing_missiles.py
```diff
@@ -109,12 +109,22 @@ def draw_entities(self, screen):
         pygame.draw.rect(screen, self.main_character_color_, main_character_rect)
         self.draw_friendly_projectiles(screen)
 
+    def determine_enemy_color(self, enemy):
+        enemy_type = enemy.get_type()
+        if enemy_type == "BasicGrunt":
+            return YELLOW
+        elif enemy_type == "NotSoBasicGrunt":
+            return WHITE
+
```
ExtraGravity, line 152: Nitpicky, but I think it'd be more clear to have an `else` statement here.

homing_missiles.py
```diff
@@ -109,12 +109,22 @@ def draw_entities(self, screen):
...
+        return RED
+
     def draw_target_tags(self, screen):
         for enemy in self.universe_.enemies():
+            enemy_color = self.determine_enemy_color(enemy)
```
ExtraGravity: This is something we're doing twice, but I don't have any reasonable solutions to not do it twice. We should make a note of this but not something we need to change for the prototype.

not_so_basic_grunt.py
```diff
@@ -0,0 +1,100 @@
...
+        self.ID_ = self.create_ID()
+        self.size_ = 15
+        self.health_ = 30
+        self.universe_ = universe
+        self.listeners_ = []
+        self.enemy_type_ = "NotSoBasicGrunt"
+
+        self.x_velocity = 0
+        self.y_velocity = 0.3 
+        self.x_accel = 0
```
ExtraGravity, line 77: Normally I'd say abbreviations are not good in variable names, but I think this one is good. `acceleration` is a long word.

not_so_basic_grunt.py
```diff
@@ -0,0 +1,100 @@
...
+from gameobject import GameObject
+
+
+class NotSoBasicGrunt(GameObject):
+    
+
+    def __init__(self, starting_position, universe):
+        self.position_ = starting_position
+
+        self.ID_ = self.create_ID()
```
ExtraGravity: I made a change in another pull request so to make `ID_` lower case everywhere, so `ID_` becomes `id_`. I'd recommend changing it now before facing a mysterious bug later.

not_so_basic_grunt.py
```diff
@@ -0,0 +1,100 @@
...
+        self.y_accel = 0
+
+        self.gravity_hole_1 = (500,300)
+
+    """
+    Update Functions
+    """
+
+    def update(self, events):
+        force_1 = self.calculate_force_from_(self.gravity_hole_1)
```
ExtraGravity: Either implements lists here for multiple forces and multiple gravity holes, or remove the `_1`s.

not_so_basic_grunt.py
```diff
@@ -0,0 +1,100 @@
...
+
+class NotSoBasicGrunt(GameObject):
+    
+
+    def __init__(self, starting_position, universe):
+        self.position_ = starting_position
+
+        self.ID_ = self.create_ID()
+        self.size_ = 15
+        self.health_ = 30
```
ExtraGravity, line 67: Variables should be in alphabetical order.

not_so_basic_grunt.py
```diff
@@ -0,0 +1,100 @@
...
+        force_1 = self.calculate_force_from_(self.gravity_hole_1)
+
+        self.x_accel = force_1[0]
+        self.y_accel = force_1[1]
+
+        self.x_velocity += self.x_accel
+        self.y_velocity += self.y_accel
+
+        #Space friction, the not_so_basic_grunt will eventuall be at a stanstill
+        self.x_velocity += -self.x_velocity*0.00001
```
ExtraGravity: I commented that these are magic numbers before, but I think leaving them with the comment is better actually. They'll clutter up the init function unnecessarily, and reading it now I knew what they were for. Perfect use of a comment :)

not_so_basic_grunt.py
```diff
@@ -0,0 +1,100 @@
...
+        self.y_velocity += -self.y_velocity*0.003
```
ExtraGravity: I'd say writing it 
`self.y_velocity -= self.y_velocity*0.003`
or
`self.y_velocity *= 0.997` 
would be better.

not_so_basic_grunt.py
```diff
@@ -0,0 +1,100 @@
...
+            self.health_ -= damage
+
+    def get_type(self):
+        return self.enemy_type_
+    """"
+    Private Functions
+    """
+    
+
+    def calculate_force_from_(self, gravity_hole):
```
ExtraGravity: I like the function name! But remove the trailing underscore

not_so_basic_grunt.py
```diff
@@ -0,0 +1,100 @@
...
+            x_force = -0.18
+        else:
+            x_force = 0.18
+
+        if y_distance < 0:
+            y_force = -0.18
+        else:
+            y_force = 0.18
+
+        return (x_force, y_force)
```
ExtraGravity: Brackets not necessary

not_so_basic_grunt.py
```diff
@@ -0,0 +1,100 @@
...
+    """"
+    Private Functions
+    """
+    
+
+    def calculate_force_from_(self, gravity_hole):
+        x_distance = gravity_hole[0] - self.position_[0]
+        y_distance = gravity_hole[1] - self.position_[1]
+
+        if x_distance < 0:
```
ExtraGravity, line 149: These few lines better written as:

```
x_force = 0.18
if x_distance < 0:
    x_force *= -1
```

not_so_basic_grunt.py
```diff
@@ -0,0 +1,100 @@
...
+            x_force = -0.18
```
ExtraGravity: Magic numbers here need to be fixed.

not_so_basic_grunt.py
```diff
@@ -0,0 +1,100 @@
...
+    Listener Functions
+    """
+
+    def report_destroyed(self):
+        for listener in self.listeners_:
+            listener.reported_destroyed(self)
+    
+    def register(self, listeners):
+        self.listeners_.append(listeners)
+        pass
```
ExtraGravity: Delete this `pass`

rifle.py
```diff
@@ -137,17 +137,29 @@ def draw_entities(self, screen):
...
                         (enemy.position_[0] * self.DRAWING_SCALE_ - (width/2),
                         (enemy.position_[1] * self.DRAWING_SCALE_ + enemy.size_/2) + (self.target_tag_y_spacing_ * self.DRAWING_SCALE_)))
 
+    def determine_enemy_color(self, enemy):
+        enemy_type = enemy.get_type()
+        if enemy_type == "BasicGrunt":
+            return RED
+        elif enemy_type == "NotSoBasicGrunt":
+            return WHITE
+
```
ExtraGravity, line 166: Should be a `else` here

homing_missiles.py
bseto, line 152: I was debating that myself, and I decided that if anyone was to add a new `elif` they'd have to move that between the `else` and the previous `elif`. This solution lets the person just add a new `elif` to the end without moving anything else

ExtraGravity, line 152: I still think the `else` is better but that's a fair argument, I can accept that..

homing_missiles.py
bseto: Yes, I noticed we are doing this twice too and we have to split up the logic for things. That being said, the "Targetting System" Should be a class that we inherit right? At the moment this is not the case so I had no choice but to put it in both "Targetting Systems". 

All the draw functions inside are also basically repeats too

ExtraGravity: I think the draw functions if done well won't be repeats. They are repeats right now because EVERYTHING is squares, but if say we wanted to make enemies triangles, then they would not look so similar.

not_so_basic_grunt.py
bseto: I will implement a list

not_so_basic_grunt.py
bseto: I agree. Didn't know all the syntax available to me :P

ExtraGravity:
I think the different colors for different enemies would be better done as a dict member variable rather than a function. So in the init function:

```
self.enemy_colors_ = { 'BasicGrunt' = RED,
                       'NotSoBasicGrunt' = WHITE
}
```

not_so_basic_grunt.py
ExtraGravity: You probably don't know about list comprehensions, but when you do implement this as lists, list comprehensions would be way to go. So this case would look something like:
`forces = [self.calculate_force_from(gravityhole) for gravityhole in gravityholes]

ExtraGravity:
screenshot 10
I've gotten this crash after playing a bit. Haven't diagnosed it myself.

ExtraGravity:
Lower the speed of this unit please. The word is hard to read with its current speed. Also if you could get them to shoot that'd be good, maybe a weapon delay of 200.

beaesguerra:
I think that the new_not_so_basic_grunts should also be a different color when in rifle mode (like how basic grunts are red and not so basic grunts are yellow in homing missiles)

beaesguerra:
Also, in homing missiles, it is hard to see the targeted basic grunt enemies because the enemy and the target border color are both red. Should we change the basic grunt color or the target color in homing missile?

beaesguerra:
I like how the white text of the not so basic grunts (when using homing missiles) makes it easier to read the tags while they are moving. 
However, the blue text of the not so basic grunts (when using rifle) is hard to read when the tags are moving. I think it is easier for users to read moving tags when the text is bright. Perhaps making the not so basic grunts yellow when using rifle and making the basic grunts blue when using homing missiles will help.

bseto:
Seems like a good idea @beaesguerra! :)

ExtraGravity:
Overall I didn't like the dark blue color, even for the slow moving targets, so I changed the color of the blue on rifle to be lighter.

beaesguerra:
Looks good!

PR #31, opened 2015-10-01
Basic grunt shooting

ExtraGravity:
The enemies shoot and do damage to the main character, but deleting the main character and creating a game over screen is a whole different story so I will leave that to a different pull request.

### DIFF
--- a/basic_grunt.py
+++ b/basic_grunt.py
@@ -80,7 +80,7 @@ class BasicGrunt(GameObject):
         y_distance = main_character_position[1] - self.position_[1]
         distance = (x_distance**2 + y_distance**2)**(1/2)
 
-        if distance > 600:
+        if distance > 400:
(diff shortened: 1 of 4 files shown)


gordon-cheung:
Looks good! But I don't see why would rifle.py draw the enemy projectile. Shouldn't the basic projectile be the same regardless if you are using rife or homing missiles?

ExtraGravity:
rifle.py draws the enemy projectile because it's targeting system is the player's vision. The player doesn't actually see the world, he sees what the targeting system sees. The idea is that the targeting system asks the universe for information, and displays what the universe tells it about. \

The basic projectile should kind of be the same, they do different damages. What would happen in C++ would be that they both inherit from projectile and then can set things like their damage or speed maybe. But for the purposes of this prototype, we will create a new object.

PR #32, opened 2015-10-02
Removed non alphabetical words from dictionary

bseto:


ExtraGravity:
What regex did you use to find them?

bseto:
in vim i did something like..
%s/._\W._//g or something
Basically \W means non Alphabetical, and i did anything before and after non alphabetical, replace with "", and then i used another regex to remove empty lines

ExtraGravity:
Looks good, merging.

PR #35, opened 2015-10-03
Implemented homing missile to target up to 5 enemies and shoot them 

beaesguerra:
Added draw_targets() to change the colors of the targeted enemies and target tags to the target_color_. Applied the same collision and listener functions in rifle.py to homingmissiles.py. Editted order of member variables. 

Removed use of abs() in RifleProjectile's update function. Removed unnecessary redrawing of main character in draw_friendly_projectiles function in RifleTargettingSystem.

### DIFF
--- a/homing_missiles.py
+++ b/homing_missiles.py
@@ -37,7 +37,7 @@ class HomingMissilesTargetingSystem():
         self.target_tag_y_spacing_ = 5
         self.target_tags_ = dict()
         self.targeting_terminal_ = TargetingTerminal(DRAWING_SCALE)
-        self.targetted_enemies_ = list() 
+        self.targeted_enemies_ = list() 
         self.text_antialias_ = 1
         self.ui_font_ = pygame.font.SysFont("monospace", self.font_size_*DRAWING_SCALE)
         self.word_generator_ = WordGenerator()
@@ -75,15 +75,16 @@ class HomingMissilesTargetingSystem():
                 if event.key == EVENT_KEY_ENTER:
                     target_id = self.get_target_id(self.current_text_)
                     if target_id and self.current_text_ != "fire":
-                        if len(self.targetted_enemies_) == 5:
-                            untargetted_enemy = self.targetted_enemies_.pop(0)
-                        self.targetted_enemies_.append(self.universe_.enemies_[target_id])
+                        if len(self.targeted_enemies_) == 5:
+                            untargetted_enemy = self.targeted_enemies_.pop(0)
+                        self.targeted_enemies_.append(self.universe_.enemies_[target_id])
+                        self.targeted_enemies_[len(self.targeted_enemies_)-1].register(self)
                        
-                    if len(self.targetted_enemies_) > 0 and self.current_text_ == "fire":
-                        for enemy in self.targetted_enemies_:
+                    if len(self.targeted_enemies_) > 0 and self.current_text_ == "fire":
+                        for enemy in self.targeted_enemies_:
                             homing_missiles_projectile = HomingMissilesProjectile(self.universe_.main_character_.position_, enemy, self.universe_)
                             self.universe_.create_friendly_projectile(homing_missiles_projectile)
-                        self.targetted_enemies_.clear()
+                        self.targeted_enemies_.clear()
 
                     self.current_text_ = ""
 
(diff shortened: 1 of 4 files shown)


homing_missiles.py
```diff
@@ -69,7 +71,18 @@ def update(self, events):
         for event in events:
             if event.type == pygame.KEYDOWN:
                 if event.key == EVENT_KEY_ENTER:
-                    target_location = self.get_target_position(self.current_text_)
+                    target_id = self.get_target_id(self.current_text_)
+                    if target_id and self.current_text_ != "fire":
+                        if len(self.targetted_enemies_) == 5:
+                            untargetted_enemy = self.targetted_enemies_.pop(0)
```
ExtraGravity: Good use of `pop`

homing_missiles.py
```diff
@@ -69,7 +71,18 @@ def update(self, events):
...
+                        self.targetted_enemies_.append(self.universe_.enemies_[target_id])
+                       
+                    if len(self.targetted_enemies_) > 0 and self.current_text_ == "fire":
```
ExtraGravity: So what happens if there's an enemy named fire? My solution would be to remove `fire` from our dictionary, we'll count it as a keyword for us.

homing_missiles.py
```diff
@@ -89,6 +102,9 @@ def draw(self, screen):
         self.draw_entities(screen)
         self.draw_target_tags(screen)
         self.targeting_terminal_.draw_terminal(screen)
+
```
ExtraGravity: Needs a draw enemy projectiles

homing_missiles.py
```diff
@@ -89,6 +102,9 @@ def draw(self, screen):
         self.draw_entities(screen)
         self.draw_target_tags(screen)
         self.targeting_terminal_.draw_terminal(screen)
+        
+        self.draw_friendly_projectiles(screen)
+        self.draw_targets(screen)
```
ExtraGravity, line 108: So the plan here is just to draw over them? I'd actually prefer that rather than a whole change of color, they have a border around them.

homing_missiles.py
```diff
@@ -123,21 +139,84 @@ def draw_target_tags(self, screen):
...
+        
+    def calculate_trajectory(self, initial_position, target_position):
+        x_distance = target_position[0] - initial_position[0]
+        y_distance = target_position[1] - initial_position[1]
+        distance = math.sqrt(x_distance**2 + y_distance**2)
+        
+        x_velocity = (x_distance * self.speed_) / distance
+        y_velocity = (y_distance * self.speed_) / distance
+
+        return (x_velocity, y_velocity)
```
ExtraGravity: Doesn't need brackets here

ExtraGravity:
Looks pretty good, just add in a `draw_grid` and the above comments I have are small ones.

ExtraGravity:
I think a lower speed would showcase the projectile better, that ts tracks. Also, you've picked the same color as Byron, white haha. So you can't tell that his enemies are targeted, you don't need to merge in his branch, but I don't think a whole change of color is suitable.

I think the words are too difficult for homing missile, I think a range of 5-7

ExtraGravity:
I totally accidentally pushed my merge with Byron's branch, I wanted to just test it, I'll revert it

homing_missiles.py
```diff
@@ -31,14 +33,16 @@ def __init__(self, universe, DRAWING_SCALE):
         self.ids_for_target_tags_ = dict() 
         self.main_character_color_ = GREEN
         self.projectile_color_ = WHITE 
+        self.target_color_ = RED
         self.target_tag_y_spacing_ = 5
         self.target_tags_ = dict()
         self.targeting_terminal_ = TargetingTerminal(DRAWING_SCALE)
+        self.targetted_enemies_ = list()
```
ExtraGravity: should be `targeted`

homing_missiles.py
```diff
@@ -108,6 +125,27 @@ def draw_entities(self, screen):
                                           main_character.size_ * self.DRAWING_SCALE_)
         pygame.draw.rect(screen, self.main_character_color_, main_character_rect)
         self.draw_friendly_projectiles(screen)
+        self.draw_enemy_projectiles(screen)
```
ExtraGravity: Can you change the order of entity drawing so that projectiles are first, enemies in the middle, main character last?

ExtraGravity:
So if you target some enemies in homing missile, then switch to rifle and shoot something you targeted, there is strange behaviour. Try it yourself.

ExtraGravity:
Projectiles need to delete themselves if their target is destroyed. 

I did this by selecting a target in homing missile, typing fire. Then killing the same target with rifle. The homing missile projectile would get stuck at the old location.

homing_missiles.py
```diff
@@ -123,21 +164,95 @@ def draw_target_tags(self, screen):
...
+    def check_collisions(self):
+        collisions = self.universe_.get_collisions(self)
+        for enemies in collisions:
+            enemies.take_damage(1000)
+
+        if collisions:
+            self.report_destroyed()
+
+    def update(self,events):
+        self.check_collisions()
```
ExtraGravity, line 231: I'd recommend here, for the projectile to determine whether it's target has been destroyed. If it's target has been destroyed, it should do a report_destroyed()

PR #37, opened 2015-10-05
Difficulty balancing

ExtraGravity:


beaesguerra:
Looks good. I like how the rifle always hits now. I also like how the rate of the difficulty scale in the beginning is quicker. I don't know if it is just me but when maximum difficulty was reached, the amount of enemies was still pretty overwhelming. Not sure if we wanted it to be overwhelming still at it's maximum difficulty.

ExtraGravity:
Yea I think max difficulty should be overwhelming, I kinda used myself as a gauging point, where if I could barely not keep up that would be the maximum difficulty.

PR #41, opened 2015-12-23
Score counter

ExtraGravity:


beaesguerra:
I think it would be nice to have a Game Over screen after the player loses so the player can see the score they got. Other than that, the score counting is working properly. 
Also, should the type of enemies be worth differently? For example, since the not so basic grunts are more difficult to destroy compared to the basic grunts, would it make sense for them to be worth more points when they are destroyed?

bseto:
I agree with the _Game Over_ screen that pauses with the score.

ExtraGravity:
I thought about different points for different enemies but it's a bit more work to do that and for simplicity sake I decided against it. If you all still want to have different points I can do that still though, but I want to finish the prototype quickly.

The same goes for the Game Over screen, but I'd be more inclined to add this one.

We have two options for the game over screen, the first is to make a separate game state. So in main.py, there's that loop that handles the game states, we could add another for a game over screen. This means we lose everything that was happening in the game, so all the existing projectiles and enemies would not appear in the game over screen. 

The other option (more difficult) is to have the score counter print the game over screen, pause all the updates for gameobjects, and delete the main character. This is difficult because I kind of hacked the solution to go back to the main menu once the character died.

bseto:
I think it would be okay to just delete the main character but let everything else keep running. You can just put an `if` in the main loop that checks if its time to show score screen, and itll just post the score really big on the screen and keep looping? That'd be pretty easy right?

ExtraGravity:
I think the enemies/enemy projectiles are reliant on the main character exisitng right now. I can adjust the code for enemies to not shoot, and enemy projectiles to delete themselves when the main character is not there. It's doable, but does go through more of the code.

ExtraGravity:
Please refer to pr #43

PR #45, opened 2015-12-27
Shield

beaesguerra:


ExtraGravity:
The shield is not being drawn on the correct spot
screenshot from 2015-12-26 22 23 32

This is an issue drawing scale, it looks fine when run with just `python3 main.py` but I run with `python3 main.py 2` and that causes problems.

ExtraGravity:
Also can we make it have a blue outline?

beaesguerra:
Fixed!

ExtraGravity:
Very cool, I was originally thinking a solid blue outline but this effect looks cool as well. Also cyan is much better than blue :)

PR #53, opened 2016-01-06
adjusted make-all script to exit if any of the makes fail

ExtraGravity:


beaesguerra:
Looks good! I like how you added the color so it's easy to notice if it failed or succeeded.

PR #54, opened 2016-01-06
improved testability of hitboxes

ExtraGravity:
changed is_overlapping from a static function to a member function so that it's possible to mock the function


bseto:
Looks good. Ya good move for helping us test

PR #55, opened 2016-01-06
generalized missiles to projectiles

ExtraGravity:
refactored missile launcher to have projectile listeners instead of explicitly missile  listeners


bseto:
Good idea. However, we need a way for Universe to distinguish between a friendly and enemy projectile when Universe passes the object to the Collision Handler. You mentioned a fix where we could create another listener function where it is like `notify_enemy_projectile` or `notify_friendly_projectile`. I think that is a good idea. 

Please implement the fix to the above problem, and then we can merge this. Good job and good idea!

ExtraGravity:
Instead of `notify_enemy...` or `notify_friendly...` I set an argument for the team that the projectile is affiliated with. I think this would provide greater flexibility for us if we wanted to have neutral objects or anything else. Let me know what you think.

test/src/missle_launcher_test.cpp
```diff
@@ -15,12 +17,35 @@ class MissleLauncherTest : public ::testing::Test {
     }
 };
 
+TEST(MissileLauncherTest, team_set) {
+    MissileLauncher friendly_launcher = MissileLauncher(FRIENDLY);
+    friendly_launcher.team() == FRIENDLY;
```
bseto: What does this do? Did you mean to write `if (friendly_launcher.team() == FRIENDLY)`?  and make it an assert or EXPECTEQ?

ExtraGravity: yea you're correct, that should be an expect EQ

src/universe.cpp
```diff
@@ -30,6 +30,6 @@ void remove_deleted_objects() {
 }
 
 /* listeners */
-void Universe::notify_missile_launched(Missile *missile) {
-    all_game_objects.push_back(missile);
-}
+void Universe::notify_projectile_launched(Projectile *projectile, Team team) {
+    all_game_objects.push_back(projectile);
```
bseto, line 34: This is also where you'll put the projectile into the collision handler along with the `team` associated with it right?

bseto:
I think it looks good. Hopefully everything is setup for you to do your collision handler. Good luck!

src/universe.cpp
ExtraGravity, line 34: yes!

PR #57, opened 2016-01-07
Missile launcher testing

ExtraGravity:
This should be a complete set of tests for missile launcher for now. Let me know if I should test for anything else.


enochtsang:
This should be a complete set of tests. Let me know if I should test for anything else.

test/src/missle_launcher_test.cpp
```diff
@@ -5,22 +5,68 @@
...
+    MockMissileLauncherListener* launcher_listener;
+
     virtual void SetUp() {
+        launcher = new MissileLauncher() ;
+        launcher_listener = new MockMissileLauncherListener();
+        launcher->add_listener(launcher_listener);
     }
 
     virtual void TearDown() {
+        delete launcher_listener;
```
beaesguerra, line 25: Shouldn't launcher also be deleted to prevent a potential memory leak?

test/src/missle_launcher_test.cpp
```diff
@@ -5,22 +5,68 @@
...
 };
 
-TEST(MissileLauncherTest, check_listeners_notified_onlaunch) {
-    MockMissileLauncherListener launcher_listener;
-    EXPECT_CALL(launcher_listener, notify_missile_launched(_))
-        .Times(1);
+TEST_F(MissileLauncherTest, multiple_listeners_notified) {
+    MockMissileLauncherListener launcher_listener2;
+
+    EXPECT_CALL(*launcher_listener, notify_missile_launched(_));
```
beaesguerra: To clarify, these 2 EXPECT_CALL checks are expected to fail right? Because the notify_missile_launched will only get called when a missile is created (which happens after these 2 EXPECT_CALL checks).

beaesguerra:
Looks good! 
This may seem redundant, but I think it would be a good idea to have a test case checking to make sure notify_missile_launched is called zero times when nothing happens. This would make sure that there is no unexpected calls to notify_missile_launched. This might be more important later on when making sure that only specific types of listeners would be notified by an event that has many listeners.

ExtraGravity:
Noted, will add.

EDIT:
I took a look at adding this test case, but I actually don't understand what is being tested here. How do you suggest I write the test case?

test/src/missle_launcher_test.cpp
ExtraGravity, line 25: Not sure, google test will actually tell you if you have a memory leak. I only had to delete the listener to fix my memory leak error, deleting the actual launcher was not necessary. Not sure why but I think including unnecessary code is worse.

bseto, line 25: I actually agree with Bea. Your "SetUp()" Is explicitly creating a `new MissileLauncher()` every time. However, your TearDown() is not deleting it.

ExtraGravity, line 25: I will look into why it does not report a memory leak then. But I am skeptical about deleting it because there is no memory leak error on compilation.

EDIT: 
http://stackoverflow.com/questions/34806872/googlemock-not-catching-leaked-object/34808027#34808027

beaesguerra:
There could be a test case where a listener is added and no missile is created. We would expect that there would be no calls to notify_missile_launched. 
It does seem a bit redundant though so maybe the check to make sure that there are no unexpected calls to notify_missile_launched would occur during integration/system testing?

ExtraGravity:
So googletest will actually fail the test if there are extra calls. So if notify was called more than it should've, the test would fail. I think this should be sufficient to test the intent you mentioned, what do you think?

beaesguerra:
Yes, that would take into account what I was concerned about.

PR #58, opened 2016-01-07
Gameobject test

ExtraGravity:
This should be a complete set of tests for gameobject. Let me know if I should test for anything else.

### DIFF
--- a/test/src/gameobject_test.cpp
+++ b/test/src/gameobject_test.cpp
@@ -13,7 +13,7 @@ protected:
     }
 };
 
-TEST(GameObjectTest, check_id_start_0) {
+TEST(GameObjectTest, id_start_0) {
     GameObject* gameobject = new MockGameObject();
     unsigned int expected = 0;
     unsigned int actual = gameobject->id();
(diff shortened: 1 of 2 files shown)


test/src/gameobject_test.cpp
```diff
@@ -0,0 +1,38 @@
...
+};
+
+TEST(GameObjectTest, check_id_start_0) {
+    GameObject* gameobject = new MockGameObject();
+    unsigned int expected = 0;
+    unsigned int actual = gameobject->id();
+    EXPECT_EQ(expected, actual);
+}
+
+TEST(GameObjectTest, incrementing_id) {
```
gordon-cheung, line 23: Would it be better to name this check_incrementing_id, I think just incrementing_id is too ambiguous, other than that  it seems okay to merge

ExtraGravity, line 23: Well I think tests are always "checking" something so I think writing check is unncessary? @bseto what do you think

bseto, line 23: Well.. I think for the sake of consistency with "check_id_start_0", it would make sense for "incrementing_id" to be named "check_incrementing_id"

ExtraGravity, line 23: I was planning to rename the other one, having a `check` in front of all our tests seems redundant. I  can remove the `check` off the other one in this PR though.

PR #59, opened 2016-01-07
Hitbox tests

gordon-cheung:


test/src/hitbox_test.cpp
```diff
@@ -23,3 +24,24 @@ TEST (HitboxTest, simple_hitbox_overlapping_false) {
...
+
+TEST (HitboxTest, constructor_2) {
+    SDL_Rect rect1 = {0, 0, 100, 100};
+    Hitbox hitbox1 = Hitbox(rect1);
+    bool actual, expected = true;
+    if (hitbox1.hitbox().x == 0 && hitbox1.hitbox().y == 0 && hitbox1.hitbox().w == 100 && hitbox1.hitbox().h == 100)
+    	actual = true;
+    else
+    	actual = false;
+    EXPECT_EQ(expected, actual);
```
ExtraGravity: It would probably better to have a `EXPECT_EQ` for every member. Also, you can probably keep this one more sparse without the `actual` and `expected`. I know i wrote that before, but it's not really necessary. 

So I would imagine it would be more like:

``` c++
EXPECT_EQ(hitbox1.hitbox().x, 0);
EXPECT_EQ(hitbox1.hitbox().y, 0);
...
```

test/src/hitbox_test.cpp
```diff
@@ -23,3 +24,24 @@ TEST (HitboxTest, simple_hitbox_overlapping_false) {
     bool actual = hitbox1.is_overlapping(hitbox2);
     EXPECT_EQ(expected, actual);
 }
+
+TEST (HitboxTest, constructor_1) {
+    Hitbox hitbox1 = Hitbox(0, 0, 100, 100);
+    bool actual, expected = true;
```
ExtraGravity: instead of keeping an expected variable for `true`. You can use 

``` c++
ASSERT_TRUE(actual);
```

test/src/hitbox_test.cpp
```diff
@@ -23,3 +24,24 @@ TEST (HitboxTest, simple_hitbox_overlapping_false) {
     bool actual = hitbox1.is_overlapping(hitbox2);
     EXPECT_EQ(expected, actual);
 }
+
+TEST (HitboxTest, constructor_1) {
```
ExtraGravity: The names could be a bit more descriptive. So instead of `constructor_1` which doesn't really say which constructor, you could write `member_constructor` and `sdl_rect_constructor`.

test/src/hitbox_test.cpp
```diff
@@ -23,3 +24,24 @@ TEST (HitboxTest, simple_hitbox_overlapping_false) {
     bool actual = hitbox1.is_overlapping(hitbox2);
     EXPECT_EQ(expected, actual);
 }
+
+TEST (HitboxTest, constructor_1) {
+    Hitbox hitbox1 = Hitbox(0, 0, 100, 100);
```
ExtraGravity, line 29: If `x` and `y` got swapped you wouldn't know with this test case. It would be better to use all different numbers, `Hitbox(1, 2, 3, 4);` would suffice.

ExtraGravity:
An additional `is_overlapping` test case which I would suggest is checking that `is_overlapping()` returns true when a hitbox is completely enclosed in a different hitbox.

test/src/hitbox_test.cpp
```diff
@@ -23,3 +24,28 @@ TEST (HitboxTest, simple_hitbox_overlapping_false) {
     bool actual = hitbox1.is_overlapping(hitbox2);
     EXPECT_EQ(expected, actual);
 }
+
+TEST (HitboxTest, simple_hitbox_overlapping_enclosed) {
```
ExtraGravity: The simple is unnecessary otherwise this PR looks good!

gordon-cheung: Made change

PR #60, opened 2016-01-08
Gamestate handler

ExtraGravity:


src/gamestate_handler.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+void GameStateHandler::set_window(SDL_Window* window) {
+    window_ = window;
+}
+
+void GameStateHandler::start(gamestates::GameStateName first_state){
+    gamestates::GameStateName next_state = first_state;
+    std::map<gamestates::GameStateName, I_GameState*>::iterator it; 
+    while(next_state != gamestates::EXIT){
+        it = gamestates_->find(next_state);
```
bseto: `it`? what does that stand for? 

EDIT: Oh, iterator.

src/gamestate_handler.h
```diff
@@ -0,0 +1,18 @@
+#pragma once 
+#include <map>
+
+#include <SDL.h>
+#include "I_gamestate.h"
+
+class GameStateHandler {
+public:
+	GameStateHandler(std::map<gamestates::GameStateName, I_GameState*>* gamestates);
+	void set_renderer(SDL_Renderer* renderer);
```
bseto: Where is GameStateHandler in this hierarchy? Why does it need a Renderer?

src/gamestate_handler.cpp
ExtraGravity: Want me to change it?

src/gamestate_handler.h
ExtraGravity: So AFAIK, we only need one renderer in this whole program. Basically, your main right now is going to be one "state", and game states take a renderer. I know in pygame we passed one surface around to ensure one window, I was thinking the renderer would be the equivalent for SDL 2.0. I wasn't sure, so let me know how it should work. 

But either way, I would highly suggest looking at GameState.h for possibly a bit more clarification. The test should help too.

bseto:
Closing pull request until it is integrated with the actual code.
I will focus on integrating this to the existing code base.

bseto:
Okay. I ran the tests, I think this pull request was good. The gamestates were handled well and I don't think there should be any issues with it. Gave it a quick test again to see that it error'd when I gave it Level_1 as the enum. 

Merging.

PR #62, opened 2016-01-19
Reversed the order of the x/ypos and x/yvel in the create missile arguments

bseto:
This change is made because missile has the argument (pos, vel). However, `create_missile` was accidentally created with the argument (vel, pos). This is to fix it before it becomes a bigger issue.   https://github.com/enochtsang/TerminalFighter/blob/master/src/missile.h#L8 

The ./make-all passed all but one test: (I don't believe has anything to do with my changes though).
[  FAILED  ] GameObjectTest.id_start_0 (1 ms)
[ RUN      ] GameObjectTest.incrementing_id
[       OK ] GameObjectTest.incrementing_id (0 ms)
[----------] 2 tests from GameObjectTest (1 ms total)

[----------] Global test environment tear-down
[==========] 13 tests from 4 test cases ran. (1 ms total)
[  PASSED  ] 12 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] GameObjectTest.id_start_0

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -172,7 +172,7 @@ int main(int argc, char* argv[])
             x = 0;
         x += 10;
 
-        test_launcher.create_missile(x, SCREEN_WIDTH / 2, 0, -2.2);
+        test_launcher.create_missile(0, -2.2, x, SCREEN_WIDTH / 2);
 
         processEvents();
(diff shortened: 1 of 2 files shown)


ExtraGravity:
Good change, merged

PR #63, opened 2016-01-24
Change makefile display warnings

bseto:


bseto:
I need help fixing what I think is the last warning. The original warning was:
`src/main.cpp:116:12: warning: deleting object of abstract class type ‘I_GameState’ which has non-virtual destructor will cause undefined behaviour [-Wdelete-non-virtual-dtor]
     delete test_state;`

So what I did was create a virtual ~I_GameState() = 0; However, things still aren't working. U want to give it a look Enoch?

ExtraGravity:
Will do, meanwhile can you check why the travis builds are failing?

bseto:
It's failing because I tried to make the ~I_GameState()= 0; So if you fix that, you'll fix the Travis builds.

bseto:
So Enoch, if you like this you can merge. However, I have one question before you do any merge. Do you want us to use `-Werror` instead of `Wall`?
 `-Werror Make all warnings into errors.`

This would force us to have no warnings for the future. It would also make Travis fail a build if there is a warning.

ExtraGravity:
@beaesguerra  What do you think Bea? Also, review the purpose of virtual destructors before reviewing this PR, it's important :)

beaesguerra:
I think we should use -Werror because then that way we are forced to correct the warnings that could cause problems later on.

bseto:
Thanks Bea :)

PR #69, opened 2016-03-15
Keyboard listener

bseto:


bseto:
Do not merge. 

Discussion with @enochtsang:
We will create another class called "Events". Keyboard will be a listener to "Events" and capture the keyboard inputs.

ExtraGravity:
We can probably test the keyboard, but not the events manager.

PR #71, opened 2016-04-26
Keyboard listener

bseto:
@enochtsang This was done a while ago, but it just needs some tests. I'll meet up with you soon so you can show me how to create a mock for the keyboard.

---

This change is [Reviewable](https://reviewable.io/reviews/enochtsang/terminalfighter/71)


ExtraGravity:
Damn this is happening? If you are gonna keep deving, I will too.

bseto:
Lol yep. I have spare time now haha. Test should be good now. You can review this whenever you have time.

test/src/keyboard_listener_test.cpp
```diff
@@ -35,13 +38,246 @@ class KeyboardListenerTest : public ::testing::Test {
...
+    sdlevent->key.keysym.sym = SDLK_f;
+    SDL_PushEvent(sdlevent);
+    events->update();
+}
+
+TEST_F(KeyboardListenerTest, test_key_g) {
+    EXPECT_CALL(*keyboard_listener, notify_keyboard_key_pressed("g"));
+    sdlevent->key.keysym.sym = SDLK_g;
+    SDL_PushEvent(sdlevent);
+    events->update();
```
ExtraGravity: Look up parameterized tests, probably a better option here.

bseto: Ok

Makefile
```diff
@@ -1,10 +1,10 @@
 #OBJS = files to compile
-OBJS = $(wildcard src/*.cpp) 
+OBJS = $(wildcard src/*.cpp) $(wildcard src/gamestates/*.cpp) 
 
 #COMPLIE_FLAGS = additional flags
 COMPILE_FLAGS = -Werror -g -std=c++0x
 
-LINK_FLAGS = `sdl2-config --cflags` `sdl2-config --libs` -lSDL2_image -lSDL2_ttf -pthread
+LINK_FLAGS = `sdl2-config --cflags` `sdl2-config --libs` -lSDL2_image -lSDL2_ttf -pthread -I src/gamestates/ -I src/
```
ExtraGravity: `-I` flags probably should have their own `INCLUDE` variable, I thought we already had one though.

src/events.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include "events_listener.h"
+#include "game_object.h" 
+#include "graphics_handler.h"
+#include "observable.h"
+#include <SDL2/SDL.h>
+
+class Events : public GameObject, public Observable<EventsListener> {
```
enochtsang: If I'm understanding this correctly, this `Events` class is for each kind of event and isn't a single controller class, in that case non plural would be better?

enochtsang: Does an event need to be a game object? Can't events be triggered through other updates and don't need to be updated themselves.

src/events.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include "events_listener.h"
+#include "game_object.h" 
+#include "graphics_handler.h"
+#include "observable.h"
+#include <SDL2/SDL.h>
```
ExtraGravity: Let's keep third party libraries separate and at the top.

src/gamestates/test_state.cpp
```diff
@@ -2,28 +2,37 @@
 
 TestState::TestState(SDL_Renderer* renderer)
 : renderer_(renderer)
-{}
-
-TestState:: ~TestState() {}
-
+{ 
+    exit = false;
```
ExtraGravity: Initalize private members inline, like above. `renderer_(renderer)`

src/gamestates/test_state.cpp
```diff
@@ -2,28 +2,37 @@
...
+    //Render red filled quad
     int x = 0;
 
     Delay delayer(false);
     for(;;)
     {
         if (x > SCREEN_WIDTH - 1)
-            x = 0; x += 10;
+            x = 0;
+        x += 10;
```
ExtraGravity, line 31: With these 2 lines beside each other, shouldn't it just be `x = 10`?

test/src/keyboard_listener_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+using ::testing::_;
+using ::testing::Property;
+using ::testing::InSequence;
+
+class KeyboardListenerTest : public ::testing::TestWithParam<std::tuple<std::string , int>> {
+public: 
+    MockKeyboardListener *keyboard_listener;
+    Keyboard             *keyboard;
+    Events               *events;
+    SDL_Event            *sdlevent;
```
ExtraGravity: Inconsistent naming here? Also I feel like the whitespacing, not everyone has a standard tool to make it this way like in go, let's just keep it easy and separate each word by one space. Also I feel like these should be private and have a trailing underscore.

src/events.h
```diff
@@ -0,0 +1,24 @@
...
+
+class Events : public GameObject, public Observable<EventsListener> {
+
+public: 
+    Events() 
+    : GameObject(0, 0) //default coordinates.
```
ExtraGravity: Perhaps we should make a default constructor for gameobject that acts this way.

src/keyboard_listener.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include <string>
+
+class KeyboardListener{
+public:
+    virtual void notify_keyboard_key_pressed(std::string keypress) = 0;
```
ExtraGravity: Maybe it's just me, but I feel like just `keyboard_key_pressed` seems more intuitive to read, or perhaps `handle_keyboard_key_pressed`. With the `notify` at the front, it reads like the listener is committing the action of doing the notifying, when really it's being notified.

I think this issue becomes really apparent in lines 38 and 39 of `keyboard.h`

``` c++
private:
     void notify_events(SDL_Event e);
     void notify_key(std::string key); 
};
```

Even though they read the same, `notify_key` is notifying the listeners, while `notify_events` is handling incoming events as a listener.

test/src/keyboard_listener_test.cpp
```diff
@@ -0,0 +1,92 @@
...
+                            std::make_tuple("q", SDLK_q),
+                            std::make_tuple("r", SDLK_r),
+                            std::make_tuple("s", SDLK_s),
+                            std::make_tuple("t", SDLK_t),
+                            std::make_tuple("u", SDLK_u),
+                            std::make_tuple("v", SDLK_v),
+                            std::make_tuple("w", SDLK_w),
+                            std::make_tuple("x", SDLK_x),
+                            std::make_tuple("y", SDLK_y),
+                            std::make_tuple("z", SDLK_z),
```
ExtraGravity: If we're gonna go all out, should we test `0-9` also?

src/keyboard.cpp
```diff
@@ -0,0 +1,155 @@
+#include "keyboard.h"
+
+void Keyboard::notify_key(std::string key) {
+    for(KeyboardListener *listener : listeners) {
+        listener->notify_keyboard_key_pressed(key);
```
ExtraGravity: Nothing to change, but as a note, this could this be really inefficient to have one notification for each key press? What if there was some way we could call `notify_keyboard_key_pressed` one for each listener per update. But premature optimization is bad, so no need to change it now.

src/keyboard.h
```diff
@@ -0,0 +1,42 @@
...
+ *  [Button]            [Return Value]
+ *  Left arrow          LEFT         
+ *  Right arrow         RIGHT
+ *  Up arrow            UP
+ *  Down arrow          DOWN
+ *  Space Key           SPACE
+ *  Backspace           BKSPACE
+ *  Esc Key             ESC
+ *  Enter Key           ENTER
+ *  Forward Slash       SLASH
```
ExtraGravity: Perhaps we should consolidate these as const variables. Some of these can definitely be represented as chars: slash, enter, and space.

src/keyboard.h
```diff
@@ -0,0 +1,42 @@
...
+ *  
+ */
+
+class Keyboard : public EventsListener, public Observable<KeyboardListener> {
```
ExtraGravity, line 29: Keyboard is pretty general, perhaps a more precise name would be `KeyPressHandler`

src/keyboard.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include "game_object.h"
+#include "keyboard_listener.h"
+#include "observable.h"
+#include <SDL2/SDL.h>
+#include <string>
+#include "graphics_handler.h"
+#include "events_listener.h"
```
ExtraGravity, line 11: Let's organize our includes to be:
1. System libraries
2. Third party libraries
3. Project libraries

src/universe.cpp
```diff
@@ -25,11 +25,15 @@ void Universe::draw_to_screen() {
...
 
-void remove_deleted_objects() {
+void Universe::remove_deleted_objects() {
     //TODO
 }
 
+void Universe::add_events_handler(Events *events) {
+    all_game_objects.push_back(events);
+}
+
```
ExtraGravity, line 35: This isn't a part of the PR but should defs be here, how did this thing work before having this function? D:

ExtraGravity (changes requested):
Took me a little bit to wrap my head around all the changes. Perhaps in the future architectural changes like this could use an explaining comment to save time. Overall though it's well done, I think the direction this PR is taking the game is a good one :)

test/src/keyboard_listener_test.cpp
bseto: Forgot about the 0-9 lol

src/gamestates/test_state.cpp
bseto: No, it looks confusing but that `if` statement just didn't have curly braces.

ExtraGravity: Crazy, can you fix that in this PR :o

ExtraGravity (approved):

src/universe.cpp
bseto: Sorry, what do you mean?

ExtraGravity:
I dont' understand this review thing, I wanted to get like outdated diffs or whatever. Cause you fixed some of them, but it doesn't show which ones are outstanding.

src/universe.cpp
ExtraGravity: Nevermind I'm dumb

bseto:
I know what you mean. I wanted to see my outdated diffs too...

ExtraGravity:
I think in the future, I won't use the review feature, just single comments.

bseto:
I was thinking the same thing. So is this PR good? 

Edit1: Woops i haven't pushed yet.
Edit2: You can manually check all the stuff now hahaha

ExtraGravity:
"Manually check all the stuff now", disgusting haha

ExtraGravity:
I'll say it now, even though I haven't read through it for specific instances but something for you to think about. With c++11/14, we should be using smart pointers everywhere, we shouldn't need to use `new` or `delete` anywhere.

src/keyboard.h
ExtraGravity: Outstanding

src/keyboard.h
ExtraGravity: Outstanding

src/keyboard.h
ExtraGravity: Outstanding

src/gamestates/test_state.h
```diff
@@ -1,26 +1,31 @@
...
-class TestState : public I_GameState {
+class TestState : public I_GameState, public KeyboardListener, public EventsListener{
 public:
 	TestState(SDL_Renderer* renderer);
-    ~TestState();
 	gamestates::GameStateName run();
 	gamestates::GameStateName name() const;
 
+    /* listeners */
+    void handle_keyboard_key_pressed(std::string keypress);
```
ExtraGravity: I think just `handle_key_press` should be sufficient for verbosity.

ExtraGravity: Outstanding

bseto: I thought you said you were going to stop using reviews. 

EDIT: Nevermind. Even single line comments are reviews now too.

ExtraGravity: Oh man cancer, I'm so sad right now.

src/keyboard.h
bseto: Finished

src/keyboard.h
bseto: I don't really get what you mean? How do I do ascii space and enter?

src/keyboard.h
bseto: Done

PR #72, opened 2016-09-16
first attempt at graphics layering

ExtraGravity:
This change is [Reviewable](https://reviewable.io/reviews/enochtsang/terminalfighter/72)

### DIFF
--- a/src/draw_request.h
+++ b/src/draw_request.h
@@ -1,0 +1,34 @@
+#pragma once
+
+#include <SDL2/SDL.h>
+
+enum GraphicPriority {
+    OVERLAY,
+    UI,
+    FRONT,
+    MIDDLE,
+    BACK,
+    BACKGROUND
+};
+
+class DrawRequest {
+
+public:
+    explicit DrawRequest(SDL_Texture *texture,
+                SDL_Rect texture_rect)
+    :texture_(texture)
+    ,texture_rect_(texture_rect)
+    {}
+
+    SDL_Texture* texture() const {
+        return texture_;
+    }
+
+    const SDL_Rect& texture_rect() const {
+        return texture_rect_;
+    }
+
+private:
+    SDL_Texture* texture_;
+    SDL_Rect texture_rect_;
+};
--- a/src/graphics_handler.cpp
+++ b/src/graphics_handler.cpp
@@ -1,22 +1,48 @@
 #include "graphics_handler.h"
 
-GraphicsHandler::GraphicsHandler(SDL_Renderer *renderer) {
-    renderer_ = renderer;
+namespace {
+    const std::vector<GraphicPriority> DRAW_ORDER = {
+        GraphicPriority::BACKGROUND,
+        GraphicPriority::BACK,
+        GraphicPriority::MIDDLE,
+        GraphicPriority::FRONT,
+        GraphicPriority::UI,
+        GraphicPriority::OVERLAY
+    };
 }
 
-void GraphicsHandler::draw(SDL_Texture *texture, SDL_Rect texture_rect) {
-    SDL_RenderCopy(renderer_, texture, NULL, &texture_rect);
+GraphicsHandler::GraphicsHandler(SDL_Renderer *renderer)
+: renderer_(renderer)
+, draw_queue_({
+        {GraphicPriority::OVERLAY, std::vector<DrawRequest>() },
+        {GraphicPriority::UI, std::vector<DrawRequest>() },
+        {GraphicPriority::FRONT, std::vector<DrawRequest>() },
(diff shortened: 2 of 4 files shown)


src/graphics_handler.cpp
```diff
@@ -1,22 +1,48 @@
...
 }
 
-void GraphicsHandler::draw(SDL_Texture *texture, int x_pos, int y_pos) {
+void GraphicsHandler::draw(SDL_Texture *texture, int x_pos, int y_pos, GraphicPriority priority) {
     SDL_Rect texture_rect;
     SDL_QueryTexture(texture, NULL, NULL, &texture_rect.w, &texture_rect.h);
     texture_rect.x = x_pos;
     texture_rect.y = y_pos;
-    SDL_RenderCopy(renderer_, texture, NULL, &texture_rect);
+    this->draw(texture, texture_rect, priority);
```
bseto, line 35: Pretty good! I like the use of `this` here.

src/draw_request.h
```diff
@@ -0,0 +1,44 @@
+#pragma once
+
+#include <memory>
```
bseto: Where is this used?

src/missile.cpp
```diff
@@ -18,7 +18,7 @@ void Missile::draw(GraphicsHandler *graphics) {
         printf("Missile graphics were null! Setting missile graphic...\n");
         set_texture(graphics->load_image("assets/projectiles/missile.png"));
     }
-    graphics->draw(missile_texture_, (int)x_pos(), (int)y_pos());
+    graphics->draw(missile_texture_, (int)x_pos(), (int)y_pos(), GraphicPriority::MIDDLE);
```
bseto, line 21: I like the `GraphicPriority` here. But just as a note: You wanted to know what the next `TODO` should be. Maybe we should work on doing that global config file soon? 
That way we can have the `GraphicPriority::MIDDLE` be something like `GlobalConfig.get(PriorityMissile)` or something along those lines. 

That way, we'll be able to have a config file that we can edit to change all of their layers.

bseto (commented):
Pretty good so far from my first read. I will go into the details tomorrow morning!

src/draw_request.h
ExtraGravity: I was trying to use smart pointers before, but I realize now it's not necessary. I'll delete this line.

src/missile.cpp
ExtraGravity: I was thinking, that maybe we should push for something playable first. Then do a big cleanup like adding using a config file, but having something playable first might help motivate us better.

bseto: ok, that sounds good

src/draw_request.h
```diff
@@ -0,0 +1,44 @@
...
+    
+    GraphicPriority z_level() const {
+        return z_level_;
+    }
+
+
+private:
+    SDL_Texture* texture_;
+    SDL_Rect texture_rect_;
+    GraphicPriority z_level_;
```
bseto: Is there a point to store the z_level_ inside GraphicPriority?

The `void GraphicsHandler::draw(SDL_Texture *texture, int x_pos, int y_pos, GraphicPriority priority)` function has `draw_queue_.at(priority).push_back(DrawRequest(texture, texture_rect, priority));`

which stores the DrawRequest in a vector according to it's priority already right?

I'm mostly wondering if you think we will ever have to query for the z_level_ in the future.

src/graphics_handler.cpp
```diff
@@ -1,22 +1,48 @@
...
 }
 
 void GraphicsHandler::update_screen() {
+    for(auto priority: DRAW_ORDER) {
+        for(auto const& draw_request: draw_queue_[priority]) {
+            SDL_RenderCopy(renderer_, draw_request.texture(), NULL, &draw_request.texture_rect());
+        }
+        draw_queue_.at(priority) = std::vector<DrawRequest>();
```
bseto, line 43: Very good! I was looking around to make sure you were clearing the old stuff to avoid memory leaks.

bseto:
I think everything looks good. You just have to take out the `include <memory>`

src/draw_request.h
bseto: Completed.

src/draw_request.h
bseto: Completed.

bseto:
Okay. Everything is done. All you gotta do is fix conflicts. 
- [x] Remove `include <memory>` header
- [x] Remove/explain `GraphicPriority z_level_;`
- [ ] Fix conflict

EDIT: Checklists like the above aren't bad on the meantime for tracking what's done haha...
reference: https://github.com/blog/1375-task-lists-in-gfm-issues-pulls-comments

PR #77, opened 2017-05-26
Added CMake

bseto:
To run this.

1. Go to top level directory of TerminalFighter
2. create a new directory called 'build' `mkdir build`
3. run `cmake ..`
4. run `make` (you can use `make -j8` or whatever if you want parallel compiling - yes, it works)
5. binary will be located inside the build/ directory, inside src/ `TerminalFighter/build/src/`

In hindsight, I think travis will fail.


ExtraGravity:
Should fix the travis build.

ExtraGravity (changes requested):
- FindSDL2_ttf.cmake was taken from somewhere right?

- For travis, using the following wget, `https://l.messenger.com/l.php?u=https%3A%2F%2Fsourceforge.net%2Fprojects%2Fastyle%2Ffiles%2Fastyle%2Fastyle%25203.0%2Fastyle_3.0_linux.tar.gz%2Fdownload&h=ATMZImYXU7vmymhRfm0DIngfMBFkkyoei50lUJuQPU8HjwDV7g2yB3jauK7Jt2J7lp0oFImOy-n3Ozhxy9o77q8ayXz1r-eetuZLojLiybdCg4zgVGDp8n2XwJb4iHiBwAatRWGcONo`

- Lastly, I think the Gamestates should stay in it's own DIR, it'll need it's own CMakeLists.txt, and it's something we should probably learn to do now. Or we're not going to use multiple DIRS in the future just because we don't want to go through the hassle of learning how to set it up.

- README should be updated with build steps

FindSDL2_ttf.cmake
```diff
@@ -0,0 +1,98 @@
...
+        # path suffixes to search inside ENV{SDLDIR}
+        include/SDL2 include
+        PATHS ${SDL2_TTF_PATH}
+        )
+
+if (CMAKE_SIZEOF_VOID_P EQUAL 8)
+    set(VC_LIB_PATH_SUFFIX lib/x64)
+else ()
+    set(VC_LIB_PATH_SUFFIX lib/x86)
+endif ()
```
ExtraGravity, line 59: This whole file you took from somewhere else right?

README.adoc
```diff
@@ -0,0 +1,24 @@
...
+`Make -j8`
++
+. Copy the executable to the correct location
+`cp src/TerminalFighter ../`
+
+== How to Build Using `build-all.sh`
+
+. Run `build-all.sh` script
++
+`./build-all.sh`
```
ExtraGravity, line 23: You mean make-all? Well I think build-all is a better name now that we dont' use makefile directly. So renaming the mak-all script would be better.

src/gamestates/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+add_library(GamestateHandler GamestateHandler.cpp)
+add_library(MenuState MenuState.cpp)
+add_library(TestState TestState.cpp)
+
+target_link_libraries(TestState GameObjectMediator Universe CollisionDetector BasicEnemy GraphicsHandler MainCharacter MissileLauncher Events Delay Keyboard)
```
ExtraGravity: Can we multiline this?

src/gamestates/I_Gamestate.h
```diff
@@ -1,6 +1,6 @@
 #pragma once
 
-#include <SDL.h>
+/* #include <SDL.h> */
```
ExtraGravity: Delete?

src/gamestates/TestState.cpp
```diff
@@ -2,12 +2,12 @@
 
 #include "TestState.h"
 
-#include "GameObjectMediator.h"
-#include "Universe.h"
-#include "CollisionDetector.h"
-#include "BasicEnemy.h"
-#include "GraphicsHandler.h"
-#include "MainCharacter.h"
+#include "src/GameObjectMediator.h"
```
ExtraGravity, line 5: All this src/ is pretty ugly, can we get rid of it?

src/gamestates/TestState.h
```diff
@@ -1,13 +1,14 @@
...
 #include <SDL2/SDL.h>
 
-#include "I_Gamestate.h"
-#include "Delay.h"
-#include "GameConstants.h"
-#include "KeyboardListener.h"
-#include "Keyboard.h"
-#include "MissileLauncher.h"
-#include "Events.h"
+#include <src/gamestates/I_Gamestate.h>
```
ExtraGravity: Use quotes instead of brackets.

test/TestAll.cpp
```diff
@@ -0,0 +1,13 @@
+#include <gtest/gtest.h>
+#include "CollisionDetectorTest.cpp"
+#include "GameObjectTest.cpp"
+#include "GameStateHandlertest.cpp"
+#include "HitboxTest.cpp"
+#include "KeyboardListenerTest.cpp"
+#include "MissileLauncherTest.cpp"
```
ExtraGravity, line 7: You shouldn't ever need to include cpp files, that's a recipe for circular dependencies. Also, gtest is supposed ot have auotomatic test discovery that doesn't need this.

ExtraGravity (changes requested):
Still need to test it on my machine.

FindSDL2_ttf.cmake
bseto: Yep

src/gamestates/CMakeLists.txt
bseto: Good idea

README.adoc
bseto: Ok, I can do that

src/gamestates/TestState.cpp
bseto: Not sure :/, tried to look into it b4 but things I tried didn't work. Tried a big combination of things with the cmake and the file includes. Il give it one more shot though

ExtraGravity: We can run the build within the src directory too, and the make all script will just cd into the src before running it's current stuff.

test/TestAll.cpp
bseto: W/o the includes, i get:
[==========] Running 0 tests from 0 test cases.
[==========] 0 tests from 0 test cases ran. (0 ms total)
[  PASSED  ] 0 tests.

Not sure why, but when we used a normal makefile, it'd discover the tests, but w/ CMake, it seems like i need the includes for these. 

Example where they do it: https://www.eriksmistad.no/getting-started-with-google-test-on-ubuntu/

src/gamestates/TestState.cpp
bseto: ? sorry i don't follow.

test/TestAll.cpp
ExtraGravity: I think there's a call in CMakeLists to add tests. https://cmake.org/cmake/help/v3.0/module/FindGTest.html I would try that, including cpp files can make some really funky errors :/

PR #80, opened 2017-06-27
rename build-all.sh to make-all.sh for better completion

ExtraGravity:


bseto:
On #77, you said that build-all would be better than make-all. 

Also, if you want it to be make-all, you are going to have to change the readme first.

ExtraGravity:
Yea tru, build-all is better. I did this cause I vaguely remember me saying make-all would be better. But no you're right haha.

PR #84, opened 2017-07-10
Bseto/create terminal

bseto:


src/I_GraphicsHandler.h
```diff
@@ -10,6 +10,7 @@ class I_GraphicsHandler {
     virtual void init() = 0;
     virtual void draw(SDL_Texture* texture, SDL_Rect texture_rect, GraphicPriority priority) = 0;
     virtual void draw(SDL_Texture* texture, int x_pos, int y_pos, GraphicPriority priority) = 0;
+    virtual void draw(SDL_Surface* surface, int x_pos, int y_pos, GraphicPriority priority) = 0;
```
ExtraGravity, line 13: What's the difference with texture?

src/Terminal.cpp
```diff
@@ -0,0 +1,62 @@
...
+}
+
+const I_Hitbox& Terminal::hitbox() const {
+    return hitbox_;
+}
+
+void Terminal::update() {
+    //do nothing - TODO: maybe add cool effects one day
+}
+void Terminal::notify_collision(GameObject& collided_object) {
```
ExtraGravity, line 38: Feels wrong that a terminal would have this as a possible function. I'll think about it.

src/Terminal.cpp
```diff
@@ -0,0 +1,62 @@
...
+    //do nothing
+}
+void Terminal::set_texture(SDL_Texture* texture) {
+    terminal_texture_ = texture;
+}
+
+void Terminal::handle_key_press(const std::string& keypress) {
+    if (keypress.length() == 1) {
```
ExtraGravity, line 46: What about shift.

src/Terminal.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include "Terminal.h"
+#include "GraphicsHandler.h"
+
+SDL_Texture* Terminal::terminal_texture_ = NULL;
+
+Terminal::Terminal(double x_pos, double y_pos, double width, double height)
+    : GameObject(x_pos, y_pos, 0.0, 0.0)
+    , hitbox_(Hitbox(x_pos, y_pos, width, height)) {
+    player_text_ = "";
+    default_font_ = TTF_OpenFont("/usr/share/fonts/truetype/ubuntu-font-family/Ubuntu-B.ttf", 84);
```
ExtraGravity: We need to think of some other way to pass this in.

src/Terminal.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include "Terminal.h"
+#include "GraphicsHandler.h"
+
+SDL_Texture* Terminal::terminal_texture_ = NULL;
+
+Terminal::Terminal(double x_pos, double y_pos, double width, double height)
+    : GameObject(x_pos, y_pos, 0.0, 0.0)
+    , hitbox_(Hitbox(x_pos, y_pos, width, height)) {
+    player_text_ = "";
```
ExtraGravity: These members variables should be instantiated inline like the above 2.

ExtraGravity (commented):
I don't think the xcf file is necessary.

Otherwise this is nice stuff, perfectly done with the testing.

src/I_GraphicsHandler.h
bseto: So with the Terminal, I get the TTF_Font stuff as a surface, and the only way to convert a surface to a texture is if I have access to the renderer. However, we have kept the renderer hidden away in GraphicsHandler so that only graphics handler will do the heavy work. 

That's why I needed graphics handler to accept a surface, and then graphics handler will convert the surface into a texture.

src/Terminal.cpp
bseto: haha yea I agree. I'm guessing through constructors + .ini file or something?

src/Terminal.cpp
bseto: It has to have this function atm because I made Terminal inherit GameObject. 

I think i should actually have it under I_Updatable (which is game_services_ inside Universe) rather than GameObject. 

^I'll get this done b4 we merge

EDIT: 
Actually, I didn't need the "Updatable" part of the GameObject, i needed the "Draw" from GameObject, which I_Updatable does not have.

src/Terminal.cpp
bseto: I think this will be ok because it'll be up to Keyboard to handle Shift + Keys (which it doesn't do atm). But when I do change the keyboard, it'll give the string as a capital letter, or whatever is supposed to happen.

PR #86, opened 2017-07-13
have missile launcher be a part of the main character

ExtraGravity:


src/gamestates/TestState.cpp
```diff
@@ -16,25 +16,34 @@ TestState::TestState(SDL_Renderer& renderer)
...
     GraphicsHandler graphics_handler(renderer_);
-    graphics_handler.init();
+    Universe universe(graphics_handler);
     std::unique_ptr<Events> events(new Events());
     std::unique_ptr<I_CollisionDetector> collision_detector(new CollisionDetector());
-    Universe universe = Universe(graphics_handler);
     GameObjectMediator game_object_mediator(universe, *collision_detector);
+    Keyboard keyboard;
 
+    // Setup engine critical components
```
bseto, line 27: I like this

bseto:
I like how it tidies up gamestates a bit. I also like how this will go for the future

bseto (changes requested):
I think my only concern is:

How will we change weapons with this configuration? and what if the user wants only one weapon active at a time (which i think is our normal use-case).

src/MainCharacter.cpp
```diff
@@ -9,10 +9,24 @@ SDL_Texture* MainCharacter::texture_ = NULL;
...
+}
+
+void MainCharacter::add_weapon(std::unique_ptr<GameObject> weapon) {
+    weapons_.push_back(std::move(weapon));
+}
+
+// GameObject functions
+
 void MainCharacter::update() {
+    for (auto& weapon : weapons_) {
```
bseto, line 27: I think we need an "active" weapon or something instead of updating them all

ExtraGravity: I don't think so, because even unequipped weapons may have a cooldown time.

ExtraGravity:
I think in the future, the MainCharacter can have an "active_weapon" variable that relates to `weapons_`, And weapons will have an "activate()" function. Either way, I think these are problems that we can solve in the future.

src/gamestates/TestState.cpp
ExtraGravity: I'm thinking we can bake this into a gamestate later. Like does a gamestate really need that much control of whether a universe and etc are created or not? I don't think so, they can just have access functions.

PR #87, opened 2017-07-15
Bseto/targeting system

bseto:


src/CMakeLists.txt
```diff
@@ -24,6 +25,7 @@ target_link_libraries(MissileLauncher Missile)
 target_link_libraries(MissileLauncher GameObjectMediator)
 target_link_libraries(Universe GraphicsHandler GameObject)
 target_link_libraries(Terminal ${SDL2_LIBRARIES} ${SDL2_TTF_LIBRARIES} GameObject Hitbox GraphicsHandler)
+target_link_libraries(TargetingSystem GameObject)
```
ExtraGravity, line 28: This shit is weird af.

src/TargetingSystem.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+void TargetingSystem::setup_local_dict(std::string relative_path) {
+
+    std::ifstream infile("assets/dictionary.txt");
+
+    //local_dict_ is 2D vector. 1st dimension is used to split words on length
+    //where index represents word length, 2nd dimension holds the words
+    local_dict_.resize(30);
+
+    for (std::string line; std::getline(infile, line); ) {
```
ExtraGravity, line 22: std::getline? cool, go doesn't even have that ;D

src/TargetingSystem.cpp
```diff
@@ -0,0 +1,69 @@
+#include "TargetingSystem.h"
+
+
+TargetingSystem::TargetingSystem(int word_length_lb, int word_length_ub, std::string color_hex)
+    : word_length_lb_(word_length_lb)
+    , word_length_ub_(word_length_ub)
+    , color_hex_(color_hex) {
+    setup_local_dict("placeholder");
```
ExtraGravity: Can you pass in "assets/dictionary.txt", less to change later and actually use `relative_path` arg in the func.

src/TargetingSystem.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+    for (std::string line; std::getline(infile, line); ) {
+        if (line.length() < 30) {
+            local_dict_[line.length()].push_back(line);
+        }
+    }
+
+    infile.close();
+
+    std::printf("%s\n", grab_word(1, 7).c_str());
```
ExtraGravity: This is extra debug right

src/TargetingSystem.cpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/*if lb is same or greater than ub, than lb will be used and ub will be ignored*/
+std::string TargetingSystem::grab_word(int lb, int ub) {
+    srand(time(NULL)); //this actually makes rand() be random
```
ExtraGravity: For future reference, dont' think we need any fancy tricks like that. We only need true randomness when dealing with security, which for this game I don't think is necessary.

src/TargetingSystem.cpp
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+
+    infile.close();
+
+    std::printf("%s\n", grab_word(1, 7).c_str());
+}
+
+/*if lb is same or greater than ub, than lb will be used and ub will be ignored*/
+std::string TargetingSystem::grab_word(int lb, int ub) {
```
ExtraGravity: Can you expand all the lb and ub in this file. I think upper_bound and stuff makes it a lot more fluent to read even though it's alot longer.

src/TargetingSystem.cpp
```diff
@@ -0,0 +1,69 @@
...
+const I_Hitbox& TargetingSystem::hitbox() const {
+
+}
+
+void TargetingSystem::notify_collision(GameObject& collided_object) {
+
+}
+
+
+void TargetingSystem::debug_vector() {
```
ExtraGravity: debug_vector doesn't really say what this does. Naming it print_dict is think is way better.

src/TargetingSystem.cpp
```diff
@@ -0,0 +1,69 @@
...
+    for (int i = 0; i < local_dict_.size(); i++) {
+        printf("%s %d\n", "Word of size:", i);
```
ExtraGravity: Don't feel like this print is necessary and can get in the way of parsing the output systematically if we ever need to do that.

src/TargetingSystem.h
```diff
@@ -0,0 +1,37 @@
...
+
+#include "GameObject.h"
+
+
+#include <string>
+#include <map>
+#include <vector>
+#include <fstream>
+#include <cstdlib>
+#include <ctime>
```
ExtraGravity: I feel like some of these can be moved to the cpp file. It's better to have includes in the cpp files to speed up compile times.

ExtraGravity (changes requested):

src/TargetingSystem.cpp
bseto: yea, was just extra debug

src/TargetingSystem.cpp
bseto: well, i wanted this cause if you run ./TerminalFighter and say the random number is... "4", then whenever you run ./TerminalFighter, the first number is always going to be 4. The sequence may be random, but it will ALWAYS be the same sequence unless you do srand(time(NULL))

PR #89, opened 2017-07-31
Bseto/targeting system

bseto:


make-all.sh
```diff
@@ -23,6 +23,6 @@ else
     printf "\n${RED}|*******| SOURCE COMPILATION FAILED |*******|${NC}\n\n"
     exit 1
 fi
-# cp src/TerminalFighter ../
+cp src/TerminalFighter/TerminalFighter ../
```
ExtraGravity, line 26: Why is this a thing?

src/Engine/GameObject/GameObject.cpp
```diff
@@ -5,20 +5,23 @@ unsigned int GameObject::id_counter_ = 0;
 GameObject::GameObject()
     : x_pos_(0)
     , y_pos_(0)
-    , id_(id_counter_++) {
+    , id_(id_counter_++)
+    , health_(0) {
```
ExtraGravity: This seems... dangerous lol.  Let's make it 1.

src/TerminalFighter/Missile/Missile.cpp
```diff
@@ -36,6 +37,7 @@ const I_Hitbox& Missile::hitbox() const {
 }
 
 void Missile::notify_collision(GameObject& collided_object) {
+    collided_object.take_damage(5); //some made up numbers. Maybe these should be part of an INI file that we can easily edit later
```
ExtraGravity, line 40: For sure will be in configuration later.

src/TerminalFighter/MissileLauncher/MissileLauncher.cpp
```diff
@@ -7,15 +7,17 @@ MissileLauncher::MissileLauncher(Team::Team team, I_GameObjectMediator& game_obj
     : team_(team)
     , game_object_mediator_(game_object_mediator)
     , terminal_((SCREEN_WIDTH / 2) - 400, SCREEN_HEIGHT - 150, 100, 30)
+    , targeting_system_(new TargetingSystem(3, 5, "FFFF"))
```
ExtraGravity: Last one is a color code right? Use FFF to be more standard. Usually hex color codes are either 3 letters or 6 letters. If it's 3, each letter gets repeated to form the 6. 

i.e. FEA -> FFEEAA

src/TerminalFighter/TargetingSystem/TargetingSystem.cpp
```diff
@@ -2,13 +2,20 @@
...
 
+#include <SDL_ttf.h>
+
 #include "TargetingSystem.h"
+#include "GraphicsHandler/I_GraphicsHandler.h"
 
 TargetingSystem::TargetingSystem(int word_length_lower_bound, int word_length_upper_bound, std::string color_hex)
     : word_length_lower_bound_(word_length_lower_bound)
     , word_length_upper_bound_(word_length_upper_bound)
+    , hitbox_(Hitbox(0, 0, 1920, 1080)) //hardcoded numbers, TODO, don't have these hardcoded
```
ExtraGravity: GameConstants.h should have screen size values.

src/TerminalFighter/TargetingSystem/TargetingSystem.cpp
```diff
@@ -2,13 +2,20 @@
...
     , color_hex_(color_hex) {
     setup_local_dict("assets/dictionary.txt");
+    srand(time(NULL)); //TODO this should be called at a higher level, maybe Universe //this actually makes rand() be random
```
ExtraGravity: Since this is bieng used everywehre, can we just call it in main? Or maybe the engine init i dunno.

src/TerminalFighter/TargetingSystem/TargetingSystem.cpp
```diff
@@ -2,13 +2,20 @@
...
+    default_font_ = TTF_OpenFont("/usr/share/fonts/truetype/ubuntu-font-family/Ubuntu-B.ttf", 24);
+    white_ = {255, 255, 255};
```
ExtraGravity: I don't feel like this needs to be a member variable, more like a constant. This shouldn't change?

src/TerminalFighter/TargetingSystem/TargetingSystem.cpp
```diff
@@ -29,32 +36,64 @@ void TargetingSystem::setup_local_dict(std::string relative_path) {
...
+        random_word_length = word_length_lower_bound_ +
+                             (rand() % (word_length_upper_bound_ - word_length_lower_bound_));
     }
 
     return local_dict_[random_word_length].at(rand() % local_dict_[random_word_length].size());
 }
 
 void TargetingSystem::update() {
-
+    printf("Checking for heartbeats...\n");
```
ExtraGravity, line 58: I'll add a logger soon lol.

src/Engine/GameObject/GameObject.cpp
```diff
@@ -27,6 +30,10 @@ void GameObject::notify_destroyed() {
     }
 }
 
+void GameObject::take_damage(int damage) {
+    health_ -= damage;
```
ExtraGravity, line 34: I think notify_destroyed if health < 0 should be the default behaviour. How targetting system is doing things is not the normal use case and should be the one overriding this function, not basic enemy or main character.

ExtraGravity (changes requested):
Lots of compiler warnings, please fix them.

make-all.sh
bseto: So with your new changes to src/ directories. the executable used to build on...
`build/src/<here>(TerminalFighter)`
But now, it builds in
`build/src/TerminalFighter/<here>(TerminalFighter)`

On your old PR, we both missed it, but the ./make-all script was actually not copying over the executable. I realized this when I noticed my changes would affect the compile, but it would not affect the actual executable.

ExtraGravity: Oh I see, I dunno why when I read this I thought it was copying  a ton of source files. This makes sense.

src/TerminalFighter/TargetingSystem/TargetingSystem.cpp
bseto: yea, but not sure where to put it. Maybe i'll read up on where people usually put it.

ExtraGravity: Use an anonymous name space at the top of the file. Lines 8-13 https://github.com/UCSolarCarTeam/Recruit-Resources/blob/master/Advanced-Recruit-Training/src/LogFileReader.cpp

bseto:
Is this good to go now?

PR #90, opened 2017-08-16
Enoch/logging

ExtraGravity:


src/TerminalFighter/main.cpp
```diff
@@ -103,14 +100,16 @@ void close() {
 }
 
 int main(int argc, char* argv[]) {
+    FLAGS_log_dir = "/tmp/TerminalFighter";
```
bseto: I kept getting a an error:
`COULD NOT CREATE A LOGGINGFILE 20170816-121402.8765!Could not create logging file: No such file or directory`
until i created `/TerminalFighter` within my `/tmp` directory. 

I think the directory should be created automatically, either by the setup/installer scripts, or have the game check if it exists or something

bseto (changes requested):

bseto:
This looks good! However, I am not getting any output from "TargetingSystem.cpp" where it should be logging to `INFO`

Example, line 67 of `TargetingSystem.cpp` should be outputting "status: alive" to the INFO log.
I am getting a log that has 18 lines from main.cpp, and then after that, i have like 80k empty lines.

Is this happening for you too?

bseto (approved):

PR #95, opened 2017-09-25
Enoch/mac build

ExtraGravity:
- Added some mac dependencies to installer.sh, most of our stuff worked out of the box. But rereading it, now that we're using cmake we should probably install SDL via cmake.

- Load in fonts locally from assets/fonts

- Fonts in Delay, Terminal, and TargettingSystem are now closed in their destructors and an error message is logged if they weren't loaded properly.

- Changed main to create the /tmp/TerminalFighter/ logging directory on mac as well as linux. Previously only linux.


src/TerminalFighter/main.cpp
```diff
@@ -61,7 +61,21 @@ bool init_SDL() {
     }
 
     //Creates the SDL Window
-    window = SDL_CreateWindow("Video Application", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, SCREEN_WIDTH, SCREEN_HEIGHT, SDL_WINDOW_SHOWN);
+#ifdef __APPLE__
+    window = SDL_CreateWindow("Video Application",
+                              SDL_WINDOWPOS_UNDEFINED,
```
bseto: Why don't we just accept an argument for this height and width?

ctlai95:
I couldn't run the make-all.sh script because I don't have `astyle` on my mac. Should we add a section to install it automatically?

ExtraGravity:
Er, I'd say it's pretty easy to install and because it's not an actual build dependency for TerminalFighter, it shouldn't be in the install script. Up for debate, but I vote no.

PR #98, opened 2017-09-29
Enoch/config

ExtraGravity:
Currently there's no safe way to load in a setting. If you do the path wrong the whole thing crashes. I have plans to add in a safe way to load in settings but I'll put that in another pull request.
 
Part of this pull request is integrating the settings, i.e. parametrizing things, so I ended up parameterizing things not related to settings as well.


bseto:
A couple things. 
You need to provide the Findyaml-cpp.cmake thing, just like how I included "FindSDL2_ttf.cmake". 
```
CMake Error at CMakeLists.txt:13 (find_package):
  By not providing "Findyaml-cpp.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "yaml-cpp",
  but CMake did not find one.

  Could not find a package configuration file provided by "yaml-cpp" with any
  of the following names:

    yaml-cppConfig.cmake
    yaml-cpp-config.cmake

  Add the installation prefix of "yaml-cpp" to CMAKE_PREFIX_PATH or set
  "yaml-cpp_DIR" to a directory containing one of the above files.  If
  "yaml-cpp" provides a separate development package or SDK, be sure it has
  been installed.


-- Configuring incomplete, errors occurred!
```

Another thing, I don't think you included the directory Settings? so I can't really find `settings.cpp`

- [x] Include the Findyaml.cpp.cmake
- [x] Include the settings directory

src/TerminalFighter/MissileLauncher/MissileLauncher.cpp
```diff
@@ -4,12 +4,30 @@
...
-MissileLauncher::MissileLauncher(Team::Team team, I_GameObjectMediator& game_object_mediator)
-    : team_(team)
+MissileLauncher::MissileLauncher(
+    Team::Team team,
+    I_GameObjectMediator& game_object_mediator,
+    std::string missile_graphic_path,
+    std::string terminal_graphic_path,
+    std::string terminal_font_path,
+    std::string targeting_system_font_path,
+    std::string targeting_system_dict_path,
```
bseto, line 14: Just curious, where do we draw the line on who gets passed "const I_Settings& settings" and who gets strings?

src/TerminalFighter/TargetingSystem/TargetingSystem.cpp
```diff
@@ -35,21 +40,26 @@ TargetingSystem::~TargetingSystem() {
     }
 }
 
-void TargetingSystem::setup_local_dict(std::string relative_path) {
-
+bool TargetingSystem::setup_local_dict(std::string relative_path) {
     std::ifstream infile(relative_path);
 
+    if (!infile) {
```
bseto, line 46: Thanks xD

src/TerminalFighter/TestState/TestState.cpp
```diff
@@ -9,28 +10,40 @@
...
 }
 
 gamestates::GameStateName TestState::run() {
+    int screen_width = settings_.video_settings()["window"]["width"].as<int>();
+    int screen_height = settings_.video_settings()["window"]["height"].as<int>();
+
+    std::vector<std::string> graphic_paths;
+    auto graphics_node = settings_.asset_paths()["graphics"];
+
+    for (YAML::const_iterator it = graphics_node.begin(); it != graphics_node.end(); ++it) {
```
bseto: It's funny you used `auto` on line 31 `auto graphics_node`, but not here where it is easy to tell `it` is an iterator ;D. 

I think for readability, line 31 should have it's typed specified, but here is a good case of using `auto` for the iterator. 

https://stackoverflow.com/a/6435066

bseto (changes requested):
I don't think there's any other problems. 
Just uhh, do you think we'll have to just decide when objects get access to Settings, and when other objects specify what settings are for them?

Like obvious case, Missile does not get the `Settings` object because we can have different missile launchers shoot different missiles. 

And then MissileLauncher doesn't get `Settings` either cause u can have different types of MissileLaunchers too and etc. Are we going to have a guideline on this or just keep it to w/e the dev wants going forward?

src/TerminalFighter/MissileLauncher/MissileLauncher.cpp
ExtraGravity: Pass in as little as you can, general programming rule. Feels different for settings, but doesn't need to be.

Issue #94, opened 2017-09-23
Audio Handler


bseto:
Using SDLMixer to play sound chunks and music

bseto:
Closing as this applied to the c++ version

PR #99, opened 2017-10-02
Bseto/audio handler

bseto:
In reference to #94


config/asset_paths.yml
```diff
@@ -3,10 +3,10 @@ fonts:
   default: "assets/fonts/Ubuntu-B.ttf"
 
 graphics:
-  missile: "assets/projectiles/missile.png"
-  terminal: "assets/terminal/futureui1.png"
-  main_character: "assets/ships/Arman.png"
-  basic_enemy: "assets/ships/BasicEnemy.png"
+  missile: "assets/images/projectiles/missile.png"
```
ExtraGravity: 👍

installer.sh
```diff
@@ -95,6 +95,9 @@ wget https://www.libsdl.org/projects/SDL_mixer/release/SDL2_mixer-2.0.0.tar.gz
 tar xzvf SDL2_mixer-2.0.0.tar.gz
 cd SDL2_mixer-2.0.0
 
+#mixer1?
+sudo apt-get install libsdl-mixer1.2-dev
```
ExtraGravity: There's a brew mac command `brew install sdl_mixer`

Also there's a SDL Mixer 2.0, why aren't we using the newer version?

src/Engine/AudioHandler/AudioHandler.cpp
```diff
@@ -0,0 +1,53 @@
...
+    if ( ret == NULL ) {
+        LOG(FATAL) << "Fatal error, could not load chunk: " << path.c_str() <<
+                   "Error: " << Mix_GetError() << " Exiting...";
+    }
+
+    sound_effects_[path.c_str()] = ret;
+
+    return ret;
+}
+void AudioHandler::play_music(Mix_Music* music) {
```
ExtraGravity: This PR doesn't include music, can we keep these music funcs out for now?

src/Engine/AudioHandler/AudioHandler.h
```diff
@@ -0,0 +1,19 @@
...
+    ~AudioHandler();
+    AudioHandler();
+    void init();
+    Mix_Music* load_music(std::string path);
+    Mix_Chunk* load_chunk(std::string path);
+    void play_music(Mix_Music* music);
+    void play_chunk(Mix_Chunk* chunk);
+
+private:
+    std::map <std::string, Mix_Chunk*> sound_effects_;
```
ExtraGravity: We keep a string map for images too, I wonder if it'd be better if we used a large enum instead of strings. That way if we type the audio name wrong, it'll be a compile error instead of a runtime error.

src/TerminalFighter/CMakeLists.txt
```diff
@@ -12,15 +12,15 @@ add_library(MenuState MenuState/MenuState.cpp)
 add_library(TestState TestState/TestState.cpp)
 add_library(Settings Settings/Settings.cpp)
 
-target_link_libraries(BasicEnemy Hitbox)
+target_link_libraries(BasicEnemy Hitbox GameObject)
```
ExtraGravity: I feel like this isn't necessary.

src/TerminalFighter/MissileLauncher/MissileLauncher.cpp
```diff
@@ -41,6 +41,7 @@ void MissileLauncher::create_missile(double x_pos, double y_pos, double x_vel, d
     std::unique_ptr<Missile> missile(
         new Missile(x_pos, y_pos, x_vel, y_vel, 5, missile_graphic_path_));
     game_object_mediator_.add_projectile(team_, std::move(missile));
+    game_object_mediator_.play_sound("assets/sounds/NFF-laser-gun-03.wav");
```
ExtraGravity: Should be loaded from settings.

src/TerminalFighter/main.cpp
```diff
@@ -12,6 +12,7 @@ extern "C" {
 }
 #include <SDL_ttf.h>
 #include <SDL2/SDL.h>
+//#include <SDL2/SDL_mixer.h>
```
ExtraGravity: Remove

test/mocks/MockGameObjectMediator.h
```diff
@@ -14,6 +16,11 @@ class MockGameObjectMediator : public I_GameObjectMediator {
         add_projectile_proxy(team, uPtr.get());
     }
 
+    virtual void play_sound(std::string path) {
```
ExtraGravity: Use MockMethod, the call would be `MOCK_METHOD1(play_sound, void(std::string);` The number after MOCK_METHOD is the number of args.

src/Engine/GameObjectMediator/I_GameObjectMediator.h
```diff
@@ -10,4 +11,5 @@ class I_GameObjectMediator {
     virtual ~I_GameObjectMediator() {}
     virtual void add_game_object(Team::Team team, std::unique_ptr<GameObject> game_object) = 0;
     virtual void add_projectile(Team::Team team, std::unique_ptr<GameObject> projectile) = 0;
+    virtual void play_sound(std::string path) = 0;
```
ExtraGravity: This should be a const function, promises that this func won't change the object.

src/Engine/AudioHandler/AudioHandler.h
```diff
@@ -0,0 +1,19 @@
...
+#include "I_AudioHandler.h"
+
+class AudioHandler : public I_AudioHandler {
+public:
+    ~AudioHandler();
+    AudioHandler();
+    void init();
+    Mix_Music* load_music(std::string path);
+    Mix_Chunk* load_chunk(std::string path);
+    void play_music(Mix_Music* music);
```
ExtraGravity: These should be const functions.

src/Engine/AudioHandler/AudioHandler.cpp
```diff
@@ -0,0 +1,53 @@
+#include <glog/logging.h>
+
+#include <SDL.h>
+#include <stdio.h>
+#include <string>
```
ExtraGravity: System libraries at the top

src/Engine/AudioHandler/AudioHandler.cpp
```diff
@@ -0,0 +1,53 @@
...
+                   "Error: " << Mix_GetError() << " Exiting...";
+    }
+
+    return ret;
+}
+Mix_Chunk* AudioHandler:: load_chunk(std::string path) {
+    Mix_Chunk* ret = Mix_LoadWAV(path.c_str());
+
+    if ( ret == NULL ) {
+        LOG(FATAL) << "Fatal error, could not load chunk: " << path.c_str() <<
```
ExtraGravity: This only for initialization?

src/Engine/AudioHandler/AudioHandler.cpp
```diff
@@ -0,0 +1,53 @@
...
+    //// Mix_Music *music; // I assume this has been loaded already
+    if (Mix_PlayMusic(music, -1) == -1) {
+        printf("Mix_PlayMusic: %s\n", Mix_GetError());
+        // well, there's no music, but most games don't break without music...
+    }
+}
+void AudioHandler::play_chunk(Mix_Chunk* chunk) {
+    if (Mix_PlayChannel(-1, chunk, 0) == -1) {
+        LOG(ERROR) << "Mix_PlayChannel: " << Mix_GetError();
+        printf("What\n");
```
ExtraGravity: Legacy

src/TerminalFighter/main.cpp
```diff
@@ -58,6 +59,15 @@ void print_renderer_info() {
     }
 }
 
+bool init_mixer() {
+    if ( Mix_OpenAudio( 44100, MIX_DEFAULT_FORMAT, 2, 2048 ) < 0 ) {
+        printf( "SDL_mixer could not initialize! SDL_mixer Error: %s\n", Mix_GetError() );
```
ExtraGravity: Use log

ExtraGravity (changes requested):

installer.sh
bseto: Sorry, I had this in here because it was asking for mixer1.2, but then I realized it was asking for it because it was trying to use it - which ties into the big bug I had before we had on messenger chat.

I'll remove this libsdl-mixer1.2

src/Engine/AudioHandler/AudioHandler.cpp
bseto: Ah, I did this wrong. Good catch. I should check if the chunk has been already loaded into the map first, and then if it isn't, then call `Mix_LoadWAV`. 

I think I had it like this because I was desperate to just get the thing working lol

src/Engine/AudioHandler/AudioHandler.h
bseto: That's a good idea, but don't we define the names in the config - so they'll only be read runtime anyways ;O.

installer.sh
ExtraGravity: So we are using sdl_mixer2?

src/Engine/AudioHandler/AudioHandler.h
ExtraGravity: Future work for me.

installer.sh
bseto: yep

src/TerminalFighter/CMakeLists.txt
bseto: lol, when linker stuff wasn't working, i was just like putting anything in anywehre trying to make it work ;D

src/Engine/AudioHandler/AudioHandler.cpp
```diff
@@ -0,0 +1,60 @@
...
+AudioHandler::AudioHandler() {
+
+}
+AudioHandler::~AudioHandler() {
+}
+void AudioHandler::init() {
+    //TODO only load really important things that need early access Like... Splash screen music or w/e
+}
+Mix_Music* AudioHandler:: load_music(std::string path) {
+    LOG(FATAL) << "Function should not be used yet";
```
ExtraGravity: :+1: really good use of fatal haha

src/Engine/AudioHandler/AudioHandler.cpp
```diff
@@ -0,0 +1,60 @@
...
+    Mix_Chunk* ret = NULL;
+    auto iterator = sound_effects_.find(path);
+
+    if ( iterator != sound_effects_.end()) {
+        ret = iterator->second;
+    } else {
+        ret = Mix_LoadWAV(path.c_str());
+
+        if ( ret == NULL ) {
+            LOG(FATAL) << "Fatal error, could not load chunk: " << path.c_str() <<
```
ExtraGravity: Small thing, but `<<` at the end of this line on the next line looks nicer.

```
cout << "blah"
     << "blah";
```

src/Engine/AudioHandler/AudioHandler.cpp
```diff
@@ -0,0 +1,60 @@
...
+}
+void AudioHandler::play_music(Mix_Music* music) const {
+    LOG(FATAL) << "Function should not be used yet";
+    //TODO just keeping code here
+    //if (Mix_PlayMusic(music, -1) == -1) {
+    //printf("Mix_PlayMusic: %s\n", Mix_GetError());
+    //}
+}
+void AudioHandler::play_chunk(Mix_Chunk* chunk) const {
+    if (Mix_PlayChannel(-1, chunk, 0) == -1) {
```
ExtraGravity: What's the 0 and -1? I think would make this more readable to encapsulte those in variables so I know what they mean without looking up SDL_Mixer docs.

src/Engine/AudioHandler/AudioHandler.h
```diff
@@ -0,0 +1,19 @@
...
+#include <map>
+
+#include "I_AudioHandler.h"
+
+class AudioHandler : public I_AudioHandler {
+public:
+    ~AudioHandler();
+    AudioHandler();
+    void init();
+    Mix_Music* load_music(std::string path);
```
ExtraGravity: Can we make these smart unique pointers?

src/Engine/GameObject/GameObject.cpp
```diff
@@ -23,7 +23,7 @@ GameObject::GameObject(double x_pos, double y_pos, double x_vel, double y_vel, i
 }
 
 GameObject::~GameObject() {
-    LOG(INFO) << "GameObject with id: " << id_ << " died";
+    //LOG(INFO) << "GameObject with id: " << id_ << " died";
```
ExtraGravity: Instead of commenting out, maybe put it as debug.

src/Engine/GameObject/GameObject.h
```diff
@@ -19,8 +22,10 @@ class GameObject : public I_Updatable, public Observable<DestroyedListener> {
     virtual void draw(I_GraphicsHandler& graphics) = 0;
     virtual const I_Hitbox& hitbox() const = 0;
     virtual void notify_collision(GameObject& collided_object) = 0;
+
     virtual void take_damage(int damage);
 
+    std::string get_sound();
```
ExtraGravity: This used anywhere?

src/Engine/GameObjectMediator/GameObjectMediator.cpp
```diff
@@ -1,11 +1,15 @@
+#include <glog/logging.h>
+
 #include "GameObjectMediator.h"
 
 #include "CollisionDetector/I_CollisionDetector.h"
 #include "Universe/I_Universe.h"
+#include "AudioHandler/AudioHandler.h"
 
-GameObjectMediator::GameObjectMediator(I_Universe& universe, I_CollisionDetector& collision_detector)
+GameObjectMediator::GameObjectMediator(I_Universe& universe, I_CollisionDetector& collision_detector, I_AudioHandler& audio)
```
ExtraGravity: I think it's about time to make this a multiline thing.

src/Engine/GameObjectMediator/GameObjectMediator.cpp
```diff
@@ -21,3 +25,14 @@ void GameObjectMediator::add_projectile(Team::Team team, std::unique_ptr<GameObj
...
 }
+
+void GameObjectMediator::play_sound(std::string path) const {
+    Mix_Chunk* sound = audio_.load_chunk(path);
+
+    if (sound == NULL) {
+        LOG(ERROR) << "Could not play chunk " << path;
+    }
+
+    printf("Playing Sound\n");
```
ExtraGravity: logger
 Styling with Markdown is supported Add review comment Cancel

src/TerminalFighter/MissileLauncher/MissileLauncher.cpp
```diff
@@ -12,11 +12,13 @@ MissileLauncher::MissileLauncher(
     std::string terminal_font_path,
     std::string targeting_system_font_path,
     std::string targeting_system_dict_path,
+    std::string missile_sound_path,
     double x_pos, double y_pos)
     : GameObject(x_pos, y_pos, 0, 0, 1)
     , team_(team)
     , game_object_mediator_(game_object_mediator)
     , missile_graphic_path_(missile_graphic_path)
+    , missile_sound_path_(missile_sound_path)
```
ExtraGravity: Better to be more specific here I think, `missile_launch_sound_path`

ExtraGravity (changes requested):
lookin good

src/Engine/AudioHandler/AudioHandler.cpp
bseto: oh ok. Yea i forgot what the 0 does, but the first argument is channel.

You can select a channel from 0-however many chanels we set (which is 4 i think). But if we don't want to choose a certain channel, we can just do -1 to let it automatically decide.

But yea i'll make it more readable

ExtraGravity: I think since the config stuff is loaded, we should add be able to make that configurable now.

src/Engine/AudioHandler/AudioHandler.h
bseto: This is actually a good idea. However I'll just take out load_music and make it a private function, and have my 
`std::map <std::string, Mix_Chunk*> sound_effects_;`
turn into 
`std::map <std::string, std::unique_ptr<Mix_Chunk*>> sound_effects_;`

That way, no function will ever need to have the Mix_Chunk. They just call `play_chunk`, and then I'll handle the rest internally and return errors accordingly. Thoughts? @enochtsang

src/Engine/GameObject/GameObject.h
bseto: no. I think I had plans for this before but not anymore

src/Engine/GameObject/GameObject.cpp
bseto: http://rpg.ifi.uzh.ch/docs/glog.html
There is no debug severity level :(

bseto: Oh just kidding. 

DLOG(INFO) i think...

src/Engine/AudioHandler/AudioHandler.h
ExtraGravity: If other things call `play_chunk` what will they pass in to specify what chunk? A path?

bseto: yea, they specify a path

ExtraGravity:
Fix da merge conflictings

src/Engine/AudioHandler/AudioHandler.h
bseto: TODO: Tried to make this work today by keeping load and play. I think it's best if i just delete "load", and have the audio_handler::play() check if it exists and loads it if it doesn't. This avoids passing the unique_ptr from function to function.

src/Engine/GameObject/GameObject.cpp
```diff
@@ -23,7 +23,7 @@ GameObject::GameObject(double x_pos, double y_pos, double x_vel, double y_vel, i
 }
 
 GameObject::~GameObject() {
-    LOG(INFO) << "GameObject with id: " << id_ << " died";
+    DLOG(INFO) << "GameObject with id: " << id_ << " died";
```
ExtraGravity, line 26: what's this?

ExtraGravity (approved):

src/Engine/GameObject/GameObject.cpp
bseto: I remember you suggested before that you wanted some stuff in DEBUG. well `LOG` doesn't have debug, you have to call `DLOG`

PR #102, opened 2017-10-05
Enoch/safe config load

ExtraGravity:
Example of safe loading

```c++
    if (!settings_.load_str(SettingsSection::ASSET_PATHS, {"fonts", "default"}, default_font_path)) {
        LOG(FATAL) << "Failed to load fonts in TestState";
    }

```


src/TerminalFighter/main.cpp
```diff
@@ -135,9 +138,15 @@ int main(int argc, char* argv[]) {
     google::InitGoogleLogging(argv[0]);
     LOG(INFO) << "Logging Intialized INFO";
 
-    bool high_dpi = settings.video_settings()["high_dpi"].as<bool>();
-    int screen_width = settings.video_settings()["window"]["width"].as<int>();
-    int screen_height = settings.video_settings()["window"]["height"].as<int>();
+    bool high_dpi = false;
+    int screen_width = 0;
+    int screen_height = 0;
```
bseto: Just curious, is there a point to having these start with a default value? If the settings don't load, we're just terminating the program so these don't actually have to start off with any value right?

bseto (commented):
Mostly took a brief look at it. Didn't have my Linux for this so I'll have to look at it again when I can have some autocompletes and editor tools like "GoTo" helping me.

src/TerminalFighter/Settings/Settings.cpp
```diff
@@ -1,44 +1,232 @@
...
+    SettingsSection section,
+    std::vector<std::string> keys,
+    bool& value) const {
+
+    YAML::Node node = load_node(section, keys);
+
+    if (!node.IsScalar()) {
+        LOG(ERROR) << "SettingsGroup " << vec_to_str(keys)
+                   << " from section " << static_cast<char>(section)
+                   << " is not boolean";
```
bseto, line 134: I understand that this message is "technically" true, but it still feels weird to me XD. 
Technically, a `string` or an `int` can get by here since they're also `scalar`. 

But I honestly have like no good suggestions so this is probs the way to go atm lol

src/TerminalFighter/Settings/Settings.cpp
```diff
@@ -1,44 +1,232 @@
...
+    errors += reload_setting(SettingsSection::VIDEO_SETTINGS);
+    errors += reload_setting(SettingsSection::ASSET_PATHS);
+
+    return !errors;
+}
+
+bool Settings::load_str(
+    SettingsSection section,
+    std::vector<std::string> keys,
+    std::string& value) const {
```
bseto, line 40: So these `load_*` functions seems pretty solid. However seems like there's lots of repeat code.
Why not - for the scalar functions - use templates <T> :3? 

That way I can use it like:
```
int screen_height;
load_setting<int>(SettingsSection::VIDEO_SETTINGS,{"window"},{"height"}, screen_height);
```

src/TerminalFighter/Settings/Settings.cpp
```diff
@@ -1,44 +1,232 @@
...
+            return false;
+        }
+    }
+
+    return true;
 }
+
+/* PRIVATE FUNCTIONS */
+
+int Settings::reload_setting(SettingsSection section) {
```
bseto, line 188: I like this. Gives the person making the GUI configuration options on how to reload settings

bseto (commented):
Couldn't get through all of it. Looks solid so far, but looks like you can make use of Templates for some of the functions.

src/TerminalFighter/main.cpp
ExtraGravity: Better to just do it safely? Good habit to initialize variables.

src/TerminalFighter/Settings/Settings.cpp
ExtraGravity: Discussed impossible

bseto (commented):
Sorry for the long delay. 

Firstly, I remember we had a discussion about detecting whether or not config/ exists. If it doesn't, we should create config/ and copy the config.example/ into it. 
Did we want to do this? 

Which brings me to my 2nd point. I believe inside config.example/ the `default` `font` is still using
`Ubuntu-b.ttf` rather than `Ubuntu-B.ttf`
EDIT: Here https://github.com/JeLLyNinjas/TerminalFighter/blob/5656023acac1d16687b9ed7c798df0cc2a85f470/config.example/asset_paths.yml#L3

src/TerminalFighter/Settings/Settings.cpp
```diff
@@ -1,44 +1,232 @@
...
+    }
+
+    LOG(FATAL) << "Attempted to load unknown settings section" << static_cast<char>(section);
+
+    // should never get here
+    return &video_settings_;
+}
+
+Settings::SettingsGroup* Settings::section_to_group(SettingsSection section) {
+    return const_cast<Settings::SettingsGroup*>(static_cast<const Settings*>(this)->section_to_group(section));
```
bseto, line 220: can you explain to me what is going on here lol...

bseto (commented):
I feel like other than some things I don't understand, it's pretty good. 
 
*Not counting if we decide to make this PR look for an existing config/ directory

src/TerminalFighter/Settings/Settings.cpp
ExtraGravity: `section_to_group` is overloaded:

```
    const SettingsGroup* section_to_group(SettingsSection section) const;
    SettingsGroup* section_to_group(SettingsSection section);
```
So default logic is to return the const version, that section of code does some nice magic to return it as non const. So now `Settings` has a const and non-const version of `section_to_group`.

ExtraGravity:
Adding copying config.example to config if it doesn't exist. Depends on `std::experimental_filesystem`, added that to build system, moved us up to c++17 in the process.

.travis.yml
```diff
@@ -16,7 +16,7 @@ addons:
       - g++-6
 
 install:
-    - g++ -v
+    - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-6 90
```
bseto: whats the 90 for :3?

bseto:
Ok, it all looks good to me. But how come I'm getting debug messages straight to my stdout now?

.travis.yml
ExtraGravity: don't know actually, copy pasted from somewhere

bseto:
logging isn't going to the /tmp/TerminalFighter/ anymore

bseto:
Completes #93

PR #104, opened 2017-10-29
Removes apt-get call from installer.sh

mpminardi:
Made it so freetype is installed by downloading / making it as not all linux distros use apt.

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -31,6 +31,7 @@ function finish {
     rm $DIR/SDL2_ttf-2.0.12.tar.gz || true
     rm $DIR/SDL2_image-2.0.0.tar.gz || true
     rm $DIR/SDL2_mixer-2.0.0.tar.gz || true
+    rm $DIR/astyle_3.0.1.tar.gz || true
     rm -rf $DIR/pkg-config-0.28 || true
     rm -rf $DIR/SDL2_image-2.0.0  || true
     rm -rf $DIR/freetype-2.8 || true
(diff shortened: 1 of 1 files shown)


bseto:
Looks good to me @enochtsang , but I have no computer to test with.

enochtsang:
I didn't use apt-get because the the apt-get astyle version was too old and didn't support `--dry-run`

installer.sh
```diff
@@ -39,10 +40,23 @@ function finish {
     rm -rf $DIR/SDL2_mixer-2.0.0 || true
     rm -rf $DIR/glog || true
     rm -rf $DIR/yaml-cpp || true
+    rm -rf $DIR/astyle || true
 }
 
 trap finish EXIT
 
+#artistic style
```
ExtraGravity: The reason this wasn't included was because this was more of a runtime installer script, not a developer one, like this one doesn't include installing cmake and stuff too. I think I would still like to keep this one as just a runtime install script.

I can see the motivation to automate installing astyle though, maybe that can be in a separate script, so for installing we keep a dev installer, runtime installer, and another one that installs for all (runs both the dev and runtime script). 

I do plan to dockerize the build process though, so at that maybe we don't even need a dev install script.  Thoughts?

ExtraGravity (commented):
The apt-get call removal is a good one, we already had code to install from source, using that universally would've been the smart thing to do :)

installer.sh
bseto: Kinda makes sense. Is there anything else thats non-dev ATM?

I was thinking this of more of a Kickstarter for devs, so we devs don't have to look for packages to download. So either we do one for runtime and then one for dev, or just put it all together (kickstart for devs) so we should include things like cmake and astyle too.

ExtraGravity: Let's make a dev and runtime installer. Dev needs all the runtime libraries, so dev use the runtime installer for one of the steps.

mpminardi: Want me to do that on this branch or as a separate pr

ExtraGravity: I think for this PR removing the apt call is good, but not adding the astyle install. So for this PR let's just do the apt-get removal, then open a separate ticket for the dev dependencies.

Issue #92, opened 2017-09-23
Menu State


bseto:
We need a Menu State to kick the game off. 

Look into how TestState is created, and create a Menu State that has a basic menu with possibly an option to launch us into the current TestState

bseto:
Closing as this applied to the c++ version

PR #105, opened 2017-11-10
Displaying the title in the menu screen

ctlai95:
MenuState renders a new MainMenu object which will hold everything in
the main menu, including the Title.

References: #92


src/TerminalFighter/MainMenu/MainMenu.cpp
```diff
@@ -0,0 +1,17 @@
+#include <glog/logging.h>
+
+#include <fstream>
+#include <string>
+#include <iostream>
+#include "MainMenu.h"
```
ExtraGravity: Order of includes is 

system
third party
internal

src/TerminalFighter/MainMenu/MainMenu.cpp
```diff
@@ -0,0 +1,17 @@
+#include <glog/logging.h>
+
+#include <fstream>
+#include <string>
+#include <iostream>
```
ExtraGravity: Seems like few of these system includes are being used.

src/TerminalFighter/Title/Title.cpp
```diff
@@ -0,0 +1,24 @@
+#include "Title.h"
```
ExtraGravity: It's unlikely we'll keep this class in the future, not generic enough, let's rename it to text box.

src/TerminalFighter/Title/Title.h
```diff
@@ -0,0 +1,17 @@
+#include <string>
+#include <SDL_ttf.h>
+
+class Title {
+    std::string title_str;
```
ExtraGravity: The order of members in this class is super weird. Should match the rest of the project.

src/TerminalFighter/Title/Title.h
```diff
@@ -0,0 +1,17 @@
...
+    SDL_Color color;
+    int x_pos;
+    int y_pos;
+    int width;
+    int height;
+    TTF_Font* font;
+public:
+    Title() : title_str(), color(), x_pos(0), y_pos(0), width(0), height(0), font() {};
```
ExtraGravity: Only the prototype should go in the header, this is an implementation. I kinda doesn't make sense to this class with default parameters, so IMO we should just forbid the default constructor altogether.

src/TerminalFighter/MenuState/MenuState.cpp
```diff
@@ -59,7 +66,10 @@ void MenuState::display_debug_frames(Delay* delayer) {
     Message_rect.w = 200; // controls the width of the rect
     Message_rect.h = 70; // controls the height of the rect
     SDL_RenderCopy(&renderer_, frame_rate_texture, NULL, &Message_rect);
-    SDL_FreeSurface(frame_rate_surface);
 }
 
-
+void MenuState::render() {
```
ExtraGravity: The graphics handler handles code like this, we should use it.

ExtraGravity (changes requested):

ExtraGravity:
Overall, nicely done. I'm proud of you :')

src/TerminalFighter/TextArea/TextArea.cpp
```diff
@@ -0,0 +1,24 @@
...
+    , width_(w)
+    , height_(h)
+    , font_(TTF_OpenFont(font_path.c_str(), font_size)) {
+}
+
+TextArea::~TextArea() {}
+
+void TextArea::render(SDL_Renderer& renderer_) {
+    SDL_Surface* surface = TTF_RenderText_Solid(font_, text_str_.c_str(), color_);
+    SDL_Texture* Message = SDL_CreateTextureFromSurface(&renderer_, surface);
```
bseto: Memory leak in this function.

You create a Surface, and then pass it into the Create Texture function - which is good. But then you don't delete the surface here. This surface will never be freed :(. gg

`SDL_FreeSurface(surface);`

src/TerminalFighter/TextArea/TextArea.cpp
```diff
@@ -0,0 +1,24 @@
...
+    SDL_Rect Message_rect;
+    Message_rect.x = x_pos_;
+    Message_rect.y = y_pos_;
+    Message_rect.w = width_;
+    Message_rect.h = height_;
+    SDL_RenderCopy(&renderer_, Message, NULL, &Message_rect);
```
bseto: SDL_Texture is also created in dynamic memory. After SDL_RenderCopying it, you can safely delete it:

`SDL_DestroyTexture(Message);`

Note you have to clean this up all manually when you do SDL_RenderCopy by yourself.

When you use the GraphicsHandler, it will do this all for you:

This is an example of it being used:
https://github.com/JeLLyNinjas/TerminalFighter/blob/5709f99229da6e750ebdf0f829e7eafae1df5abd/src/TerminalFighter/Missile/Missile.cpp#L30

And here is the line where graphics handler does cleanup:
https://github.com/JeLLyNinjas/TerminalFighter/blob/5709f99229da6e750ebdf0f829e7eafae1df5abd/src/Engine/GraphicsHandler/GraphicsHandler.cpp#L73

In hindsight.. there is no comments on Graphics handler yet (in progress). I will make sure to finish this right now lol...

bseto (changes requested):
Memory Leak!

bseto:
Not sure where, but you still have a memory leak.
If you just hang out on the menu screen, and take a system monitor, you'll see you are still memory leaking.

bseto:
For some reason, I am still getting the memory leak. @enochtsang, can you check?

ExtraGravity:
Nope no memory leak

bseto (approved):
Marking as approved for me. Letting the PR rest on Enoch :stuck_out_tongue:

src/TerminalFighter/MainMenu/MainMenu.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+#include <chrono>
+#include <unistd.h>
+
+#include <SDL_ttf.h>
+#include <SDL2/SDL.h>
+
+#include "../TextArea/TextArea.h"
+
+namespace{
```
ExtraGravity: I think this should public ENUM, then anyone interacting with the MainMenu can actually look at enum constants instead of ints.

src/TerminalFighter/MenuState/MenuState.h
```diff
@@ -19,6 +21,7 @@ class MenuState : public I_GameState {
 
 private:
     void display_debug_frames(Delay* delayer);
+    void render();
```
ExtraGravity: Is this being used?

src/TerminalFighter/TextArea/TextArea.h
```diff
@@ -0,0 +1,20 @@
+#include <string>
+#include <SDL_ttf.h>
```
ExtraGravity: new line between these two includes.

src/TerminalFighter/TextArea/TextArea.h
```diff
@@ -0,0 +1,20 @@
...
+namespace{
+    const SDL_Color COLOR_WHITE = {255, 255, 255};
+    const SDL_Color COLOR_GRAY = {100, 100, 100};
+}
+
+class TextArea {
+public:
+    TextArea(std::string, SDL_Color, int, int, int, int, std::string, int);
+    ~TextArea();
+    void render(SDL_Renderer& renderer);
```
ExtraGravity: Does everything else call these render?  They used `draw()` instead right?

src/TerminalFighter/TextArea/TextArea.cpp
```diff
@@ -0,0 +1,30 @@
...
+    , width_(w)
+    , height_(h)
+    , font_(TTF_OpenFont(font_path.c_str(), font_size)) {
+}
+
+TextArea::~TextArea() {}
+
+void TextArea::render(SDL_Renderer& renderer_) {
+    SDL_Surface* surface = TTF_RenderText_Solid(font_, text_str_.c_str(), color_);
+    SDL_Texture* Message = SDL_CreateTextureFromSurface(&renderer_, surface);
```
ExtraGravity: Message should be lower case.

src/TerminalFighter/TextArea/TextArea.cpp
```diff
@@ -0,0 +1,30 @@
...
+    SDL_FreeSurface(surface);
+    SDL_Rect Message_rect;
```
ExtraGravity: variable shoudl be lower case, also can this be a member variable? Like I feel like we should only have to initialize this once.

src/TerminalFighter/TextArea/TextArea.h
```diff
@@ -0,0 +1,20 @@
...
+    void set_color(SDL_Color color);
+private:
+    std::string text_str_;
+    SDL_Color color_;
+    int x_pos_, y_pos_, width_, height_;
+    TTF_Font* font_;
```
ExtraGravity: Should use a smart pointer

src/TerminalFighter/MenuState/MenuState.cpp
```diff
@@ -31,16 +43,32 @@ gamestates::GameStateName MenuState::run() {
...
                         case SDLK_RETURN:
-                            LOG(INFO) << "Enter was Pressed!";
-                            return gamestates::TEST;
+                            LOG(INFO) << "Enter was pressed!";
+                            if (main_menu.get_current_selection() == START) {
+                                LOG(INFO) << "Start selected";
+                                return gamestates::TEST;
+                            } else if (main_menu.get_current_selection() == QUIT) {
+                                LOG(INFO) << "Quit selected";
+                                exit(0);
```
ExtraGravity: Return `gamestates::EXIT` instead so that higher level code can handle a proper exit.

src/TerminalFighter/MainMenu/MainMenu.cpp
```diff
@@ -0,0 +1,75 @@
...
+    quit_.render(renderer);
+}
+
+void MainMenu::move_up_selection() {
+    if (current_selection_ == 0) {
+        current_selection_ = 1;
+    } else if (current_selection_ == 1){
+        current_selection_ = 0;
+    }
+    update_colors();
```
ExtraGravity: This should be called once by an `update()` function that the creator is responsible for calling.

src/TerminalFighter/MainMenu/MainMenu.h
```diff
@@ -0,0 +1,30 @@
...
+    MainMenu(std::string font_path, int width, int height);
+    ~MainMenu();
+    void render(SDL_Renderer& renderer);
+    void move_up_selection();
+    void move_down_selection();
+    void update_colors();
+    int get_current_selection();
+
+private:
+    TextArea title_;
```
ExtraGravity: These would be better as a map of `<std::string, TextArea>`, or even better replace std::string with the enum that I described before.

ExtraGravity (changes requested):
In my opinion the correct way to abstract this is to have this inherit a `Menu` interface. This interface is something that can be rendered and you can determine what selection is grabbed, and promises to deal with inputs. But we can cross that hurdle later.

This Menu interface would go in the Engine, while an implementation such as MainMenu would go in the actual game code.

bseto:
I like the idea of having a Menu interface. @ctlai95

bseto:
No memory leak anymore for me. I think b150756 fixed it.

src/TerminalFighter/TextArea/TextArea.h
ctlai95: Not sure what you mean here?

ExtraGravity: The other classes call it `draw()` instead of `render()` i think

ctlai95: Oh I see, yea they're called draw. I'll make the change

src/TerminalFighter/TextArea/TextArea.h
ctlai95: Need some time to figure out how to do this :/

src/TerminalFighter/TextArea/TextArea.h
```diff
@@ -0,0 +1,21 @@
+#include <string>
+
+#include <SDL_ttf.h>
+
+namespace{
+    const SDL_Color COLOR_WHITE = {255, 255, 255};
+    const SDL_Color COLOR_GRAY = {100, 100, 100};
```
ExtraGravity: We should make a different task, to have one file for colors. So to use them it'd be like `TerminalFighterColours::WHITE`

src/TerminalFighter/MainMenu/MainMenu.cpp
```diff
@@ -0,0 +1,76 @@
+#include <string>
+#include <glog/logging.h>
+
+#include "MainMenu.h"
+
+MainMenu::MainMenu(std::string font_path, int width, int height)
```
ExtraGravity: Passing width, height is so ass haha. We gotta change the grid space.

src/TerminalFighter/MainMenu/MainMenu.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+#include <chrono>
+#include <unistd.h>
+
+#include <SDL_ttf.h>
+#include <SDL2/SDL.h>
+
+#include "../TextArea/TextArea.h"
+
+enum Options { START, QUIT };
```
ExtraGravity: This should get moved into the class so it's private to the class. 

https://stackoverflow.com/questions/2503807/declaring-an-enum-within-a-class

you still put it in the public section though, but it doesn't pollute the global namespace.

src/TerminalFighter/MenuState/MenuState.cpp
```diff
@@ -15,10 +15,22 @@ gamestates::GameStateName MenuState::run() {
     if (!settings_.load_str(SettingsSection::ASSET_PATHS, {"fonts", "default"}, default_font_path)) {
         LOG(FATAL) << "Failed to load default font in MenuState";
     }
+    int window_width, window_height;
+    if (!settings_.load_int(SettingsSection::VIDEO_SETTINGS, {"window", "width"}, window_width)) {
+        LOG(FATAL) << "Failed to load window width in MenuState";
+    }
+    if (!settings_.load_int(SettingsSection::VIDEO_SETTINGS, {"window", "height"}, window_height)) {
```
ExtraGravity, line 22: We should make another namespace that is "ConfigPaths", so we get compiler errors if we type a path wrong, not a runtime error.

src/TerminalFighter/MenuState/MenuState.cpp
```diff
@@ -31,16 +43,34 @@ gamestates::GameStateName MenuState::run() {
...
+                            break;
+
+                        case SDLK_UP:
+                            main_menu.move_up_selection();
+                            main_menu.update();
+                            break;
+
+                        case SDLK_DOWN:
+                            main_menu.move_down_selection();
+                            main_menu.update();
```
ExtraGravity: `update()` should be called exactly once per loop.

src/TerminalFighter/TextArea/TextArea.h
```diff
@@ -0,0 +1,21 @@
...
+}
+
+class TextArea {
```
ExtraGravity: This class will probably be a dependency for a test one day. It should have an interface super class.

ExtraGravity (changes requested):

src/TerminalFighter/MainMenu/MainMenu.h
```diff
@@ -7,17 +7,17 @@
 
 #include "../TextArea/TextArea.h"
 
-enum Options { START, QUIT };
 
 class MainMenu {
 public:
     MainMenu(std::string font_path, int width, int height);
     ~MainMenu();
+    enum Options { START, QUIT };
```
ExtraGravity: put this at the top of the public separated by a new line. This is a pretty important piece and should be easy to spot.

src/TerminalFighter/MainMenu/MainMenu.h
```diff
@@ -0,0 +1,26 @@
...
+    ~MainMenu();
+    enum Options { START, QUIT };
+    void draw(SDL_Renderer& renderer);
+    void move_up_selection();
+    void move_down_selection();
+    void update();
+    MainMenu::Options get_current_selection();
+
+private:
+    TextArea title_;
```
ExtraGravity: These should use the I_TextArea interface. We should probably use Dependency injection so we can instantiate this class with fake text areas.

src/TerminalFighter/TextArea/I_TextArea.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+class I_TextArea {
+public:
+    virtual void draw(SDL_Renderer& renderer) = 0;
```
ExtraGravity: Add a virtual destructor.

`virtual I_TextArea() {}`, not pure.

ExtraGravity (changes requested):

src/TerminalFighter/MainMenu/MainMenu.h
ExtraGravity: Actually, fuck the DI, just make them use the interface class. This one can't be because it has to be instantiated, but you can do it for the vector of text  areas.

ctlai95: I dont get it, doesn't TextArea use the I_TextArea interface already?

ExtraGravity: The important part is that stuff using TextArea uses I_TextArea, this way if we create a class that depends on I_TextArea, we don't have to pass in an actual TextArea, we can pass in a MockTestArea (testing purposes!)

bseto: @ctlai95 , do you understand this^ We can re-phrase if needed

ctlai95: @bseto worked on it wtih enoch

ExtraGravity (approved):

PR #106, opened 2017-11-29
Tai/main menu options

ctlai95:


src/TerminalFighter/TextArea/TextArea.h
```diff
@@ -1,11 +1,15 @@
 #include <string>
 #include <SDL_ttf.h>
 
+#define WHITE {255, 255, 255}
```
ExtraGravity: Use anonymous name spaces instead of defines.

```
namespace {
    int FOO 4;
}
```

src/TerminalFighter/MainMenu/MainMenu.cpp
```diff
@@ -6,17 +6,62 @@
...
 MainMenu::~MainMenu() {}
 
 void MainMenu::render(SDL_Renderer& renderer) {
     title_.render(renderer);
+    start_.render(renderer);
+    quit_.render(renderer);
+}
+
+void MainMenu::move_up_selection() {
+    current_selection_ = (2 + (current_selection_ + 1) % 2) % 2;
```
ExtraGravity: This math isn't the easiest to reason through. Can we do this more readable with conditionals?

src/TerminalFighter/MainMenu/MainMenu.cpp
```diff
@@ -6,17 +6,62 @@
...
+        0.1 * height,
+        font_path,
+        68)
+    , quit_(
+        "Quit",
+        GRAY,
+        0.45 * width,
+        0.7 * height ,
+        0.1 * width,
+        0.1 * height,
```
ExtraGravity: Comments explaining the reasoning for these numbers would be helpful. It's not immediately clear what they do.

src/TerminalFighter/MainMenu/MainMenu.h
```diff
@@ -7,12 +7,22 @@
 
 #include "../TextArea/TextArea.h"
 
+const int START = 0;
+const int QUIT = 1;
```
ExtraGravity: Put these in an anonymous namespace.

ExtraGravity (changes requested):

bseto:
I like it. However, I feel like you can make this more generic. We will add more options in the future. I know at the moment, our config files are used for user configurations, but I think we will also one day have our own developer configs where it's like "encrypted". (So maybe have all the values of title, start, and quit in a config). 

Maybe for now, we can keep it like this as @enochtsang may argue that this is ok for the main menu.

But I feel in the future, we should have a pretty generic menu object where you can give it an array of menu options, and it'll just dynamically build the the menu. 

The options could include (as an example):
- Menu options (Start, Settings, Quit, etc..)
- Menu Border (toggle)
- Menu transparency value
- Leftmost X,Y position
- Scrollable
- Sound effects etc

bseto (approved):

ctlai95:
I really like those ideas, let's discuss it more in the future. Merging this to the main branch for now.

PR #108, opened 2017-12-02
Bseto/graphics documentation

bseto:


src/Engine/GraphicsHandler/I_GraphicsHandler.h
```diff
@@ -8,19 +8,191 @@
 class I_GraphicsHandler {
 public:
     virtual ~I_GraphicsHandler() {}
+    /**
+     * Used to intiate the GraphicsHandler
+     * Loads the vector of <code>graphic_paths</code> which are file paths
+     * to <code>GraphicsHandler::game_graphics_</code>, which is a map of std::string to SDL_Textures*
+     * @param path is the relative (recommended) or absolute path to graphics assets that will
```
ExtraGravity: why is relative recommended? 

I think the format of the comment should refer to the name of the actual arg so

```
graphics_paths is a vector of relative (recommended) or absolute paths...
```

src/Engine/GraphicsHandler/I_GraphicsHandler.h
```diff
@@ -8,19 +8,191 @@
...
 public:
     virtual ~I_GraphicsHandler() {}
+    /**
+     * Used to intiate the GraphicsHandler
+     * Loads the vector of <code>graphic_paths</code> which are file paths
+     * to <code>GraphicsHandler::game_graphics_</code>, which is a map of std::string to SDL_Textures*
+     * @param path is the relative (recommended) or absolute path to graphics assets that will 
+     *  be loaded when init is called. This initial load avoids having to load assets while the
+     *  level has started, and can be done when "loading" the level
+     * @return returns <code>void</code>. However, <code>GraphicsHandler::internal_load_image()</code> calls
```
ExtraGravity: the @return I think can be confusing, since you'er right it is void. It'd be better to put this stuff in the note.

src/Engine/GraphicsHandler/I_GraphicsHandler.h
```diff
@@ -8,19 +8,191 @@
...
+    /**
+     * Used to intiate the GraphicsHandler
+     * Loads the vector of <code>graphic_paths</code> which are file paths
+     * to <code>GraphicsHandler::game_graphics_</code>, which is a map of std::string to SDL_Textures*
+     * @param path is the relative (recommended) or absolute path to graphics assets that will 
+     *  be loaded when init is called. This initial load avoids having to load assets while the
+     *  level has started, and can be done when "loading" the level
+     * @return returns <code>void</code>. However, <code>GraphicsHandler::internal_load_image()</code> calls 
+     * <code>load_image</code> which _can_ potentially call <code>LOG(FATAL)</code>
+     * and exit the program when an incorrect path is called.
```
ExtraGravity: invalid path is more normal to say I think

src/Engine/GraphicsHandler/I_GraphicsHandler.h
```diff
@@ -8,19 +8,191 @@
...
+     * \note
+     * We have no real error handling here. If an invalid path is given, it means an object may not have
```
ExtraGravity: _There_ is no real error handling

third person

src/Engine/GraphicsHandler/I_GraphicsHandler.h
```diff
@@ -8,19 +8,191 @@
...
+     * an asset, so calling exiting the program is kind of appropriate. However, maybe one day, we should
```
ExtraGravity: what do you mean an object may not have an asset?

Also, removing the "calling" and the "kind of"

src/Engine/GraphicsHandler/I_GraphicsHandler.h
```diff
@@ -8,19 +8,191 @@
...
+     * handle it gracefully and have it load a dummy asset.
+     */
     virtual void init(const std::vector<std::string>& graphic_paths) = 0;
-    virtual void draw(SDL_Texture* texture, SDL_Rect texture_rect, GraphicPriority priority,
+    /**
+     * Draw function
+     * This draw function does not immediately draw the texture onto the screen. Instead, the draw function 
+     * will place the <code>SDL_Texture*</code> into a queue using the <code>priority</code> parameter to determine
+     * the queue order. The queue is then drawn by <code>Universe::draw_to_screen()</code>, which is called
```
ExtraGravity: I think the stuff after `which is called` is unnecessary and confusing.

src/Engine/GraphicsHandler/I_GraphicsHandler.h
```diff
@@ -8,19 +8,191 @@
...
+     * @param texture will be used to SDL_RenderCopy onto the rendering target (the <code> renderer_</code>)
+     * @param dest_rect is the <code>dstrect</code>, or destination rect on where the texture will
+     *  be copied to. This rect defines the size, and position the texture will be rendered to
+     * @param priority is used to define the order that the texture will be copied to the renderer.
+     *  The GraphicPriority is an enum found in "DrawRequest/DrawRequest.h"
+     * @param cleanup is used to define whether or not the draw function needs to clean up a texture.
+     *   Set this to false when the texture passed to this function is cached. 
+     *   Sometimes, the texture is cached by the GraphicsHandler::game_graphics_, in which case, 
+     *   you do not want to set the bool to `true` or else this function will delete the cached pointer.
+     *   Set this value to true when you need the function to cleanup the texture for you.
```
ExtraGravity: Can you provide examples in the code when this is true or false?

ExtraGravity (changes requested):

src/Engine/GraphicsHandler/I_GraphicsHandler.h
bseto: interesting. I thought it was path for some reason. 

relative because if you specify absolute path, my path to assets is: `/home/byron/Documents/TerminalFighter/assets`, while yours could be completely different.

src/Engine/GraphicsHandler/I_GraphicsHandler.h
bseto: What I meant is that if the given path is invalid, then essentially, the asset would not be loaded. Therefore the object (missile.h) for example, would not have a loaded asset (the missile.png)

bseto: I just realized that this isn't the case on init. Init will just cache things ahead of time. The objects will try to grab from a path anyways.

ExtraGravity (approved):

PR #110, opened 2017-12-05
add docs for gameobjectMediator

ExtraGravity:


bseto (approved):
Looks good to me. Built properly with Doxygen

src/Engine/GameObjectMediator/I_GameObjectMediator.h
```diff
@@ -6,10 +6,34 @@
...
+    /**
+     * This function is similar to <code>add_game_object()</code> except it's to add projectiles.
+     * The difference is that projectiles won't collide with other projectiles.
+     * Refer to <code>add_game_object()</code> for more information.
+     */
     virtual void add_projectile(Team::Team team, std::unique_ptr<GameObject> projectile) = 0;
+
+    /**
+     * This function gives access to play sounds in the sound engine.
+     * @param the relative path to the sound to be played.
```
bseto: Doxygen thinks `the` is the parameter. The first word is the parameter, so this line should be..
`@param path is the relative path to the sound to be played.`

bseto (approved):
Sorry, caught one little error. Please change this and then merge it. Keeping it as approved so you can merge after u fix.

PR #111, opened 2017-12-05
Bseto/docker documentation

bseto:


make-all.sh
```diff
@@ -27,3 +27,16 @@ fi
 cp src/TerminalFighter/TerminalFighter ../
 ./test/runTests
 cd ..
+
+docker --version
```
ExtraGravity: Remove

make-all.sh
```diff
@@ -27,3 +27,16 @@ fi
...
+docker build . -t terminal_documentation
+
+# Mount the current directory into /documentation, which already exists in the
+# Dockerfile as the WORKDIR.
+#
+# "doxygen -g" will create a default config file named Doxyfile
+# "patch < Doxyfile.patch will apply the configuration patch
+#sudo docker run --rm -v `pwd`:/documentation test1 /bin/bash -c "doxygen -g && \
+    #patch < Doxyfile.patch && doxygen Doxyfile
+sudo docker run --rm -v `pwd`:/documentation terminal_documentation /bin/bash -c "doxygen Doxyfile"
```
ExtraGravity: We shouldn't have sudo in these commands, really hurts portability. Instead the caller should be required to have root permissions.

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+# Apt packages
+RUN apt-get update
+# We don't want any interactive prompts
+RUN DEBIAN_FRONTEND=non-interactive apt-get install -yq doxygen graphviz
+
+# Set WORKDIR
+WORKDIR /documentation
+
+# Copy all files from this project root host computer to /documentation in the docker container
+ADD . /documentation
```
ExtraGravity: This creates a big dockerfile, and requires a rebuild everytime. Better to not have this and just mount the volume directly. Look at projects at CSI for example.

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ubuntu:xenial
+
+# Apt packages
+RUN apt-get update
+# We don't want any interactive prompts
+RUN DEBIAN_FRONTEND=non-interactive apt-get install -yq doxygen graphviz
```
ExtraGravity: Concatenate these run commands for a smaller image.

.gitignore
```diff
@@ -27,3 +27,4 @@ GRTAGS
 GTAGS
 compile_commands.json
 config/
+documentation/
```
ExtraGravity: What is this dir for actually?

ExtraGravity (changes requested):

make-all.sh
bseto: ok

Dockerfile
bseto: ok, so I had the mounting done right. I just thought I'd have to do a ADD (copy) too haha

.gitignore
bseto: This directory is where the documentation is going to be built. I want it to be ignored so it doesn't end up on the repository.

make-all.sh
```diff
@@ -27,3 +27,12 @@ fi
 cp src/TerminalFighter/TerminalFighter ../
 ./test/runTests
 cd ..
+
+# Update docker if there are any new files
+docker build . -t terminal_documentation
```
ExtraGravity: The image name should be a variable. And we should prefix the org name.

`IMAGE_NAME=JeLLyNinjas/TerminalFighterDocs` is my suggestion

make-all.sh
```diff
@@ -27,3 +27,12 @@ fi
...
+
+# "doxygen -g" will create a default config file named Doxyfile
+# "patch < Doxyfile.patch will apply the configuration patch
+#sudo docker run --rm -v `pwd`:/documentation test1 /bin/bash -c "doxygen -g && \
+    #patch < Doxyfile.patch && doxygen Doxyfile
+docker run --rm -v `pwd`:/documentation terminal_documentation /bin/bash -c "doxygen Doxyfile"
```
ExtraGravity: `doxygen Doxyfile` isn't complicated so we don't need to invoke it as a bash specific command.
```
docker run --rm -v `pwd`:/documentation terminal_documentation doxygen Doxyfile
```
should be fine

ExtraGravity (changes requested):

make-all.sh
```diff
@@ -4,6 +4,8 @@ RED='\033[0;31m'
 GREEN='\033[0;32m'
 NC='\033[0m'
 
+IMAGE_NAME=jellyninjas/terminalfighterdocs
```
ExtraGravity: I think using the caps I suggested would be better. Normal naming for docker containers.

ExtraGravity (changes requested):

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ubuntu:xenial as setup
+
```
ctlai95: you can probably remove this extra space

make-all.sh
```diff
@@ -27,3 +27,15 @@ fi
...
+docker build . -t terminal_documentation
+
+# Mount the current directory into /documentation, which already exists in the
+# Dockerfile as the WORKDIR.
+#
+# "doxygen -g" will create a default config file named Doxyfile
+# "patch < Doxyfile.patch will apply the configuration patch
+#sudo docker run --rm -v `pwd`:/documentation test1 /bin/bash -c "doxygen -g && \
+    #patch < Doxyfile.patch && doxygen Doxyfile
+sudo docker run --rm -v `pwd`:/documentation terminal_documentation /bin/bash -c "doxygen Doxyfile"
```
ctlai95: you should be mapping pwd/documentation:documentation right? otherwise only the contents of documentation will be mounted to pwd and your gitignore won't ignore the output

ctlai95:
Don't forget to change the permissions of the `documentation` folder to user

- [x] Change Permissions

make-all.sh
bseto: Caps don't work as Docker names

documentation/doxygen-build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+#apt-get install patch
+
+# "doxygen -g" will create a default config file named Doxyfile
+# "patch < Doxyfile.patch will apply the configuration patch
```
ctlai95: missing a `"`

documentation/doxygen-build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+#apt-get install patch
```
ctlai95: Is this line needed?

ctlai95 (changes requested):

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM ubuntu:xenial
+
+# Apt packages
+# We don't want any interactive prompts
+RUN apt-get update && DEBIAN_FRONTEND=non-interactive apt-get install -yq doxygen graphviz patch
+
+# Set WORKDIR
+WORKDIR /documentation
```
ExtraGravity: I probably want to use this file for builds later as well. Let's put this in `/TerminalFighter/Documentation`

make-all.sh
```diff
@@ -4,6 +4,7 @@ RED='\033[0;31m'
 GREEN='\033[0;32m'
 NC='\033[0m'
 
+IMAGE_NAME=jellyninjas/terminalfighterdocs
```
ExtraGravity: Same with this, so just

jellyninjas/terminalfighter

documentation/doxygen-build.sh
ExtraGravity: This comment isn't necessary right?

make-all.sh
```diff
@@ -27,3 +28,9 @@ fi
 cp src/TerminalFighter/TerminalFighter ../
 ./test/runTests
 cd ..
+
+# Update docker if there are any new files
+docker build . -t $IMAGE_NAME
+
+docker run --rm -v `pwd`:/documentation $IMAGE_NAME ./documentation/doxygen-build.sh
```
ExtraGravity: Let's rename the script to not be doxygen dependent, so just `build-docs.sh`

ExtraGravity (changes requested):

ctlai95:
we decided to remove the `experimental/filesystem` dependency right? I can't build this on my mac because I think the base branch isn't merged with master.

ExtraGravity (approved):

ctlai95 (approved):

PR #114, opened 2017-12-10
abstract out colours into TF_Colours namespace

ExtraGravity:


ctlai95 (changes requested):
I know it's my fault for using a different spelling but can you change all `color` to `colour` because :canada:

ctlai95:
Actually on second thought we should use `color` to be consistent with SDL's spelling. :us: :us: :us:

PR #115, opened 2017-12-17
Enable builds using docker

ExtraGravity:
, no runtime libraries, no travis or runtime env, no static compile


bseto:
So just some notes. 
Travis is failing. Noteworthy things:

Line 4739: cannot remove '//pkgconfig.tgz': No such file or directory

Also looks like your  `./docker_build.sh all` in the `.travis.yml` is failing with permission denied, and then tons of other permission denied.

build.sh
```diff
@@ -0,0 +1,55 @@
...
+
+readonly DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+NC='\033[0m'
+
+function src {
+	mkdir -p "$DIR"/build
+	(cd $DIR/build && cmake ..)
```
bseto, line 11: How come we are launching subshell commands?

build.sh
```diff
@@ -0,0 +1,55 @@
...
+
+function format {
+	(cd "$DIR" && astyle *.h *.cpp -r --options=astylerc)
+	find "$DIR" -name "*.orig" -delete
+}
+
+function docs {
+	(cd "$DIR"/documentation && doxygen -g)
+	(cd "$DIR"/documentation && patch --no-backup-if-mismatch < $DIR/documentation/Doxyfile.patch)
+	(cd "$DIR"/documentation && doxygen Doxyfile)
```
bseto: replace this line with:
```
(doxygen "$DIR"/documentation/Doxyfile)
```
Doxygen has to be run on the root directory so it can find "./src" (defined in the doxyfile)

Tested this on both vanilla bash (non-docker), and with docker and it works.

docker_build.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+readonly DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+readonly IMAGE_NAME="jellyninjas/terminalfighter"
+
+if [[ $# -eq 0 ]]; then
+	echo "usage: docker_build [init | build | <build.sh args>]"
```
bseto: I don't really understand the `build` option... 

I just run 

`./docker_build docs`
or
`./docker_build src`

But running the "build" option gives me the following:
```
./docker_build build
usage: ./build.sh [all | src | format | docs | clean]
```

And giving it no options is kinda confusing too:

```
sudo ./docker_build.sh
usage: docker_build [init | build | <build.sh args>]
usage: ./build.sh [all | src | format | docs | clean]
```

bseto (changes requested):
I really like this addition man. I like how it also gives me the option to run it without docker too - very very nice. 

Everything I've tested works:

Building docs (docker and non-docker with my suggestion. w/o my suggestion, it builds empty docs)
Building Executables (docker and non-docker)

However, I feel like the travis should pass. So you need to get travis to work. Probably by getting it to allow you to change permissions and stuff.

Also, one last thing.
Did you want to ship `build.sh` and `docker_build.sh` without chmod 755?

It'd be kinda annoying to have to chmod them every single time after a pull...

build.sh
ExtraGravity: When stuff has to be run in certain directories, it just makes more sense compared to changing a global state in the code.

docker_build.sh
ExtraGravity: The build option wasn't supposed to be there, usage statement is updated. 

```
	echo "usage: docker_build [init | shell | <build.sh args>]"
```

ExtraGravity:
Line 4739: cannot remove '//pkgconfig.tgz': No such file or directory 

is a warning i will ignore. It's not hard to reason why that is in the installer script, but it's ok it's non issue.

bseto (approved):

bseto:
Looks good to me

PR #118, opened 2017-12-21
Bseto/spritesheet reader

bseto:
I'm thinking it might be worth it to just leave the config part to a second review. So i'll keep this PR up, maybe give it a review, and then i'll make another PR to add functionality to read draw orders from a config file.


assets/images/credits.txt
```diff
@@ -0,0 +1,5 @@
+futureui1.png
+https://opengameart.org/content/sci-fi-ui-panel
+
+fireball effect
+https://opengameart.org/content/high-res-fire-ball
```
ExtraGravity: I don't imagine us using these in the future. So I don't think we need this file.

src/Engine/Delay/Delay.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 #include <chrono>
 #include <unistd.h>
+#include <string>
```
ExtraGravity, line 4: Necessary?

src/Engine/SpriteAnimator/SpriteAnimator.cpp
```diff
@@ -0,0 +1,70 @@
+#include "SpriteAnimator.h"
+
+SpriteAnimator::SpriteAnimator()
```
ExtraGravity: Is just `Animator` sufficient?

src/TerminalFighter/MainCharacter/MainCharacter.cpp
```diff
@@ -45,8 +45,19 @@ void MainCharacter::draw(I_GraphicsHandler& graphics) {
     for (auto& weapon : weapons_) {
         weapon->draw(graphics);
     }
-
-    graphics.draw(texture_, (int)x_pos(), (int)y_pos(), GraphicPriority::MIDDLE, true);
+    SDL_Rect src_rect;
+    SDL_QueryTexture(texture_, NULL, NULL, &src_rect.w, &src_rect.h);
+    src_rect.x = 0;
+    src_rect.y = 0;
```
ExtraGravity: This is pretty ugly for everything that needs to draw something IMO. This is worth adding an overloaded function for since Graphics Handler is capable of finding this information on its own.

src/Engine/SpriteAnimator/SpriteAnimator.cpp
```diff
@@ -0,0 +1,70 @@
...
+    : initialized_(false)
+{}
+
+// Keeping this as an init right now because the objects may not
+int SpriteAnimator::init(SDL_Texture* texture, int rows, int cols, int on_every) {
+    //If there was no draw_order given as an argument, we will create a default
+    if (draw_order_.size() == 0) {
```
ExtraGravity: Instead of having an if statement here, I think having the logic in the one with the draw_order, and this function will call the draw order func but pass in a draw order.

src/Engine/SpriteAnimator/SpriteAnimator.cpp
```diff
@@ -0,0 +1,70 @@
...
+        for (int i = 0; i < rows*cols; i++) {
+            draw_order_.push_back(i);
+        }
+    }
+    on_every_ = on_every;
```
ExtraGravity: I think `period_` is better than `on_every_`

src/Engine/SpriteAnimator/SpriteAnimator.cpp
```diff
@@ -0,0 +1,70 @@
...
+    rows_ = rows;
+    cols_ = cols;
+    //total_frames_ = rows * cols;
```
ExtraGravity: Remover this comment.

src/Engine/SpriteAnimator/SpriteAnimator.cpp
```diff
@@ -0,0 +1,70 @@
...
+    current_index_ = 0;
+    int w, h;
```
ExtraGravity: Instead of creating these vars, just pass in `total_sprite_sheet_size_.w` and same for the height.

src/Engine/SpriteAnimator/SpriteAnimator.cpp
```diff
@@ -0,0 +1,70 @@
...
+int SpriteAnimator::init(SDL_Texture* texture, int rows, int cols, int on_every) {
+    //If there was no draw_order given as an argument, we will create a default
+    if (draw_order_.size() == 0) {
+        for (int i = 0; i < rows*cols; i++) {
+            draw_order_.push_back(i);
+        }
+    }
+    on_every_ = on_every;
+    rows_ = rows;
+    cols_ = cols;
```
ExtraGravity: I like doing all this boring stuff at the top. Right now the logic is separated and harder to think through.

src/Engine/SpriteAnimator/SpriteAnimator.cpp
```diff
@@ -0,0 +1,70 @@
...
+    if (incrementor_ == 0) {
+        current_index_++;
+    }
+    current_index_ = current_index_ % draw_order_.size();
+
+    return return_rect;
+}
+
+SDL_Rect SpriteAnimator::get_next_frame() {
+    int throw_away;
```
ExtraGravity: nice

src/Engine/SpriteAnimator/SpriteAnimator.cpp
```diff
@@ -0,0 +1,70 @@
...
+    returned_frame = draw_order_.at(current_index_);
+    SDL_Rect return_rect;
+
+    int on_row = draw_order_.at(current_index_) / cols_; //integer division for the floor
+    int on_col = draw_order_.at(current_index_) % (rows_ - 1);
+
+    return_rect.w = total_sprite_sheet_size_.w / cols_;
+    return_rect.h = total_sprite_sheet_size_.h / rows_;
+    return_rect.x = on_col * (total_sprite_sheet_size_.w / cols_);
+    return_rect.y = on_row * (total_sprite_sheet_size_.h / rows_);
```
ExtraGravity: What do you think of creating a vector that precalculates all this in the init, a map of a rect to a frame? So then grabbing next frame is alot simpler.

ExtraGravity (changes requested):

src/Engine/Delay/Delay.h
bseto: yea, line 14 needs it

src/Engine/SpriteAnimator/SpriteAnimator.cpp
bseto: sorry, in what "one with the draw_order"? can you rephrase this sentence? I don't understand

src/Engine/SpriteAnimator/SpriteAnimator.cpp
bseto: I think it's ok, but it's not really an optimization, and we shouldn't optimize lots of these things for now.

The reason why it's not an optimization (even though it'll be in init), is because init happens during the gameplay time, and not during the "init" section of the states (which we don't have, other than commenting).

If the player shoots an enemy, they create a `missile`, the missile then goes through it's `init`, where it'll have to pre-calculate for ALL of it's things, right away, rather than doing a single calc every loop. So this, probably won't be considered a optimization, and if we want to optimize it, this extra work that I'll do, will be erased anyways.

So i'm opting not to do it for now because it won't be an optimization.

src/TerminalFighter/Terminal/Terminal.cpp
```diff
@@ -36,9 +36,41 @@ void Terminal::draw(I_GraphicsHandler& graphics) {
...
+        ui_src_rect.h = 0;
+    } else {
+        ui_src_rect.w = UIText->w;
+        ui_src_rect.h = UIText->h;
+    }
+
+    SDL_Rect terminal_src_rect;
+    SDL_QueryTexture(terminal_texture_, NULL, NULL, &terminal_src_rect.w, &terminal_src_rect.h);
+    terminal_src_rect.x = 0;
+    terminal_src_rect.y = 0;
```
ExtraGravity: My biggest concern is having to create this src_rect for every class. Everytime we have a new object we have this same ugly 4 lines of code.

src/TerminalFighter/Missile/Missile.cpp
```diff
@@ -33,8 +33,24 @@ void Missile::draw(I_GraphicsHandler& graphics) {
         set_texture(graphics.load_image(texture_path_));
     }
 
-    graphics.draw(missile_texture_, (int)x_pos(), (int)y_pos(), GraphicPriority::MIDDLE, true,
-                  util::angle(0, 0, x_vel_, y_vel_), NULL);
+    if (!spriteAnimator_.is_initialized()) {
+        //3 columns, 2 rows of sprite sheets, update every 6 frames
+        spriteAnimator_.init(missile_texture_, 2, 3, 3);
```
ExtraGravity, line 38: Weren't we gonna load this via config?

ExtraGravity (changes requested):

src/TerminalFighter/Missile/Missile.cpp
bseto: that was going to be a whole other PR

src/TerminalFighter/Terminal/Terminal.cpp
bseto: sorry, i did not replace this with the new fix.
The code for example on basicEnemy.cpp is:

```
    graphics.draw(texture_,
                  graphics.create_default_rect(texture_),
                  (int)x_pos(),
                  (int)y_pos(),
                  GraphicPriority::MIDDLE,
                  true,
                  0,
                  NULL);
 }
 }
```

ExtraGravity: Kk I saw this in a few places. After this is fixed we can merge.

ExtraGravity: Can we write unit tests for the animator class?

src/Engine/Animator/Animator.cpp
```diff
@@ -0,0 +1,70 @@
...
+    //If there was no draw_order given as an argument, we will create a default
+    if (draw_order_.size() == 0) {
+        for (int i = 0; i < rows * cols; i++) {
+            draw_order_.push_back(i);
+        }
+    }
+
+    // We will run off the assumption that the given
+    // spritesheet has it's starting sprites on 0,0.
+    // If there is padding... manually get rid of it for now
```
ExtraGravity: This might imply manually getting rid of it in code, I think it should be clearer that this means manually remove padding from the pictures using external tools.

src/Engine/Animator/Animator.h
```diff
@@ -0,0 +1,66 @@
...
+     */
+    Animator();
+    /**
+     * Uses the <code>SDL_Texture*</code> to create a <code>SDL_Rect</code> that represents the
+     * spritesheet.
+     */
+    int init(SDL_Texture* texture, int rows, int cols, int period);
+    /**
+     * Uses the <code>SDL_Texture*</code> to create a <code>SDL_Rect</code> that represents the
+     * spritesheet.
```
ExtraGravity: I don't think this comment is accurate. the `init` function doesn't really create anything.

src/Engine/Delay/Delay.h
ExtraGravity: But it didn't need it before?

src/Engine/GraphicsHandler/I_GraphicsHandler.h
```diff
@@ -19,10 +21,15 @@ class I_GraphicsHandler {
...
-     * There is no real error handling here. If an invalid path is given, the function will throw a fatal 
+     * There is no real error handling here. If an invalid path is given, the function will throw a fatal
      * error and then the program will exit.
      */
     virtual void init(const std::vector<std::string>& graphic_paths) = 0;
+
+    /**
+     * Used as a helper function to create a src_rect
+     */
+    virtual SDL_Rect create_default_rect(SDL_Texture* texture) = 0;
```
ExtraGravity: Kindo f a weird func, I think this func should have an example.

src/TerminalFighter/Missile/Missile.cpp
```diff
@@ -33,8 +33,24 @@ void Missile::draw(I_GraphicsHandler& graphics) {
...
+    }
+
+
+    SDL_Rect dest_rect;
+    dest_rect.x = (int)x_pos();
+    dest_rect.y = (int)y_pos();
+    dest_rect.w = 172;
+    dest_rect.h = 30;
```
ExtraGravity, line 46: Can we call the helper function here of just passing in the x and y pos.

src/TerminalFighter/TargetingSystem/TargetingSystem.cpp
```diff
@@ -102,8 +102,20 @@ void TargetingSystem::update() {
 void TargetingSystem::draw(I_GraphicsHandler& graphics) {
     for (std::map<int, GameObjectStringPair*>::iterator it = targets_.begin(); it != targets_.end(); ++it) {
         SDL_Surface* UIText = TTF_RenderText_Blended(default_font_, it->second->assigned_word_.c_str(), TF_Colors::WHITE);
-        graphics.draw(UIText, (int)it->second->game_object_.x_pos(), (int)it->second->game_object_.y_pos(),
-                      GraphicPriority::UI, true);
+        SDL_Rect ui_src_rect;
+        ui_src_rect.x = 0;
+        ui_src_rect.y = 0;
+        ui_src_rect.w = UIText->w;
+        ui_src_rect.h = UIText->h;
```
ExtraGravity: The utility func should be used here. Also why is UIText cap case? Shouldn't it be snake_case?

src/TerminalFighter/Terminal/Terminal.cpp
```diff
@@ -36,9 +36,36 @@ void Terminal::draw(I_GraphicsHandler& graphics) {
...
+    SDL_Rect ui_src_rect;
+    ui_src_rect.x = 0;
+    ui_src_rect.y = 0;
+
+    if (UIText == NULL) {
+        ui_src_rect.w = 0;
+        ui_src_rect.h = 0;
+    } else {
+        ui_src_rect.w = UIText->w;
+        ui_src_rect.h = UIText->h;
```
ExtraGravity: We should use the utility function here.

ExtraGravity (changes requested):

src/TerminalFighter/TargetingSystem/TargetingSystem.cpp
bseto: Ok, I'll have to create another utility function, probably overload it. But the original utility function took SDL_Texture as a argument, that's why I didn't use it here.

src/TerminalFighter/Missile/Missile.cpp
bseto: The other functions automatically took the input SDL_Texture, and re-used their original `w` and `h`. In this case, we don't want to re-use it because it'd be the size of the entire sprite sheet's `w` and `h`. So we have to specify the `w` and `h` by ourselves.

src/Engine/Delay/Delay.h
bseto: Yea it didn't need it before. Not sure why it needs it now.

ExtraGravity (approved):

PR #120, opened 2018-02-01
Bseto/1 1gamespace

bseto:
# 1-1 Gamespace Main Refactor Branch
This Branch will be used as the main refactor branch.
This branch changed Graphics Handler and Draw Request to support 1-1 gamespace.

## Make objects use 1-1
- [x] Tweak BasicEnemy to use 1-1
- [x] Tweak Missile to use 1-1
- [x] Tweak MainCharacter to use 1-1
- [x] Tweak Targeting System to use 1-1
- [x] Tweak Terminal to use 1-1
- [x] Graphics Handler has to adjust for angles from Gamespace to Screenspace

## Collision handler has to accept 1-1
- [ ] Re-write the algorithm
- [x] Make use of JN_Rect

## Text in 1-1 Gamespace needs a helper function/object
- [x] Think of a way to apply text with 1-1 gamespace

Notes on the Jn_Ttf (the helper function)
Jn_Ttf right now works off of SDL_TTF's output text, and scaling the original SDL_Surface to a JN_Rect making use of ratios based off it's height.

However, apparently, not all text surfaces have the same height even though they are rendered with the same "font size". Example: THIS_TEXT is larger than this_text. But because I'm working with ratios (which make the end result the same height), the sizes of the output text is slightly wonky. I'll need to look into a fix on this. 

http://www.sdltutorials.com/sdl-ttf
might need to think of font sizes in terms of %'s


ExtraGravity:
Holy fk

src/Engine/JN_TTF/JN_TTF.cpp
```diff
@@ -0,0 +1,23 @@
+#include "JN_TTF.h"
```
ExtraGravity: This should be JN_Ttf, JN_TTF looks like a constant.

ExtraGravity (changes requested):

src/Engine/JN_TTF/JN_TTF.cpp
bseto: But TTF is like SDL, it's a capcase thing... Would you still want it Ttf?

ctlai95:
damn dude

bseto:
@ctlai95 @enochtsang. 
Any of you want to implement: https://gamedevelopment.tutsplus.com/tutorials/collision-detection-using-the-separating-axis-theorem--gamedev-169?
 :D

PR #121, opened 2018-02-02
Updating the SDL2 release package links

bseto:
Ever since I moved to Ubuntu 17.10, the old packages were giving me lots of "SDL redeclared" errors. 

Updating the SDL2 packages for me fixed it. 

@enochtsang or @ctlai95, please test to see if this works on your mac/16.04LTS


ExtraGravity:
Travis is having bad times, gotta fix that first.

ctlai95:
I made some changes, mind if I push to this branch?

ExtraGravity:
@ctlai95 push away

installer.sh
```diff
@@ -2,18 +2,17 @@
...
 unameOut="$(uname -s)"
 case "${unameOut}" in
-    Linux*)     MACHINE=Linux;;
-    Darwin*)    MACHINE=Mac;;
-    CYGWIN*)    MACHINE=Cygwin;;
-    MINGW*)     MACHINE=MinGw;;
+    Linux*)     MACHINE=Linux ;;
+    Darwin*)    MACHINE=Mac ;;
+    CYGWIN*)    MACHINE=Cygwin ;;
+    MINGW*)     MACHINE=MinGw ;;
```
ExtraGravity, line 8: why

installer.sh
```diff
@@ -104,8 +107,8 @@ cd $DIR
 # glog
 git clone https://github.com/google/glog.git
 cd glog
-mkdir build
-cd build
+mkdir tmp
+cd tmp
```
ExtraGravity, line 115: what's wrong with build

installer.sh
```diff
@@ -89,13 +90,15 @@ cd $DIR
...
 ./configure
 make -j4
 make install -j4
-cd ../../
+cd $DIR
+
+#SMPEG
+wget https://www.libsdl.org/projects/smpeg/release/smpeg2-2.0.0.tar.gz
+tar xzf smpeg2-2.0.0.tar.gz
+cd smpeg2-2.0.0
```
ExtraGravity, line 104: why did this change?

installer.sh
ctlai95: Auto change from editor

installer.sh
ctlai95: On my mac it complains that build already exists

installer.sh
ctlai95: For me smpeg didn't exist in the previous section

installer.sh
ExtraGravity: instead of creating /tmp, maybe we can just cd in to build without making it.

ExtraGravity (changes requested):

installer.sh
ctlai95: TIL apple's HFS is case-insensitive by default, and that's why it's conflicting with the file `BUILD` in `glog/`

what do

installer.sh
bseto: Actually, same with me. My installer always stopped at 
```
./installer.sh: line 94: cd: external/smpeg2-2.0.0: No such file or directory
```

My usual fix was to just ignore it and comment it out, and re-run the script.

installer.sh
bseto: We could just do `mkdir -p build`.
Adding the -p should stop it from complaining if the `build/` directory already exists

ctlai95: I still get `mkdir: build: Not a directory`

ExtraGravity (approved):

PR #126, opened 2020-05-01
Refactored background rendering; Added Profiling

sam-ao:


terminalfighter.py
```diff
@@ -26,8 +33,10 @@ def terminalfighter(screen):
     spawn_controller = SpawnController(universe, score_counter)
     gamemaster = GameMaster(universe, spawn_controller, score_counter)
     # pygame ticks, one tick is 1/1000 second
-    # 15 pygame ticks per update is approximately 30 updates per second
-    FRAME_LENGTH_TICKS = 33
+    # 15 pygame ticks per update is ~60 updates per second
+    FRAME_LENGTH_TICKS = 15
```
sam-ao: Upped FPS to 60

sam-ao:
profile/terminalfighter_baseline.prof is the profile before the background rendering change, and profile/terminalfighter_1588373774.prof is the profile after the change.

profile/README.md
```diff
@@ -0,0 +1,9 @@
+Profiling
+---------
+###Prereqs
+
+`pip3 install snakeviz`
```
mpminardi: Should add this as a dev dependency via `pipenv install --dev snakeviz` so it is persisted in the repo (and then add the changes to the pipfile / the lock). Then this README text could probably be added to the root readme.

terminalfighter.py
```diff
@@ -99,11 +109,38 @@ def terminalfighter(screen):
...
+                         (GAME_WIDTH, i),
+                         line_width)
+
+    return background_surface
+
+
+if __name__ == '__main__':
+    # Start profiling
+    pr = cProfile.Profile()
+    pr.enable()
```
mpminardi: Is this something that should always be enabled? Does it have a very high overhead? Could add an environment variable or CLI argument to enable if it is.

mpminardi:
Does adding the actual binary profiling files add anything that could not be regenerated later? Might not make sense to persist them in the repo

terminalfighter.py
bseto: https://docs.python.org/3.8/library/profile.html#introduction-to-the-profilers
Apparently it has a "reasonable overhead", but I do agree, I think it might be nice to have a CLI argument to enable it, and have it off by default

something like --profile or something to enable this profile

rifle.py
```diff
@@ -95,35 +94,10 @@ def update(self, events):
...
-        self.draw_background(screen)
-        self.draw_grid(screen)
         self.draw_entities(screen)
         self.draw_target_tags(screen)
         self.targeting_terminal_.draw_terminal(screen)
 
-    def draw_background(self, screen):
-        pygame.draw.rect(screen, BLACK, pygame.Rect((0, 0), screen.get_size()))
-
-    def draw_grid(self, screen):
```
bseto, line 107: So I think you broke the main.py file (you must be running only the terminalfighter.py). The grid no longer shows up on the main menu and the bullet doesn't get erased when it travels through space

terminalfighter.py
```diff
@@ -99,11 +109,38 @@ def terminalfighter(screen):
         else:
             print("WARNING: Cannot keep up with 30FPS update time!")
 
-if __name__ == '__main__':
-    GAME_WIDTH = 1000
-    GAME_HEIGHT = 700
+def draw_background():
```
bseto: So the idea with each weapon having it's own "draw_grid" was that each weapon can determine what the "UI" looked like.

I know for rifle and homing missile we had the same "UI" which was this grid looking thing, but each weapon should be able to decide what the "UI" looks like. So I think you should put these back into the weapons, but call them as you have here, where you call draw_background() once..

Although actually, instead of once, you might have to call draw_background() everytime the player switches the weapon, since it'll draw a new overlay, meaning you'll have to re-draw the black
`pygame.draw.rect(screen, BLACK, pygame.Rect((0, 0), screen.get_size()))`

Also note that we won't always have this being black.. we may have a scrolling background in the future (but maybe it'll update at like 5fps or something) so idk... but i'll be ok for this PR to just have the backgrounds support black only, but I do think we should put this grid stuff back to the weapons so they can choose their own UI

bseto (changes requested):

profile/README.md
bseto: also to run it using pipenv..

`pipenv run snakeviz profile/<profilefilename>`

sam-ao: done

terminalfighter.py
sam-ao: done

rifle.py
sam-ao: fixed

terminalfighter.py
sam-ao: I've moved the draw_grid back into the weapons. For now, I've kept the draw background in the global terminalfighter.py though

bseto (approved):

